PR #46, opened 2026-08-25
fix(core): detect stale lockfile inputs

ujjwalredd:
## Problem

`eos_lockfile_is_current()` only compared project name, package count, package name, and requested version. Changes to project version, resolved version, source URL, checksum, or build type could therefore leave a stale lockfile marked current, weakening reproducible-build guarantees.

## Approach

- Reuse one helper for explicit checksums and generated metadata checksums.
- Compare project version and every serialized package input.
- Match entries one-to-one without depending on package order.
- Add regressions for project version, package name, requested/resolved versions, source, explicit/generated checksums, build type, and reordered entries.

Public signatures and the lockfile format are unchanged.

## Validation

- Regression before fix: **failed as expected**; changed lockfile inputs were incorrectly accepted as current.
- Release CMake build: **passed**.
- Targeted `test_config`: **passed**.
- Full CTest: **21/21 passed**.
- `python3 -m pytest tests -q`: **10 passed**.
- ASan/UBSan `test_config`: **passed**.
- `git diff --check`: **passed**.
- Independent review: **approved with no findings** after adding missing name/requested-version regression coverage.

## Additional considerations

The host Release build emits existing warnings in unrelated tests; neither changed C file adds a warning. No migration is needed because the serialized format is unchanged—existing stale lockfiles will simply be regenerated when their inputs differ.

GitHub Actions created Python, build, simulation, CodeQL, and assignment runs for this PR, but all are currently `action_required` pending maintainer approval.

### DIFF
--- a/core/src/lockfile.c
+++ b/core/src/lockfile.c
@@ -9,6 +9,20 @@
 #include <string.h>
 #include <ctype.h>
 
+static void package_hash(const EosPackageConfig *pc, char *hash) {
+    if (pc->hash[0]) {
+        strncpy(hash, pc->hash, EOS_HASH_LEN - 1);
+        hash[EOS_HASH_LEN - 1] = '\0';
+        return;
+    }
+
+    char input[2048];
+    snprintf(input, sizeof(input), "%s:%s:%s:%s",
+             pc->name, pc->version, pc->source,
+             eos_build_type_str(pc->build_type));
+    eos_cache_compute_hash(input, strlen(input), hash, EOS_HASH_LEN);
+}
+
 EosResult eos_lockfile_generate(EosLockfile *lock, const EosConfig *cfg) {
     memset(lock, 0, sizeof(*lock));
     strncpy(lock->project_name, cfg->project.name, EOS_MAX_NAME - 1);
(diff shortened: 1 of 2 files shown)


srpatcha:
### Review — verified, ready to merge (after #61)

The defect is real: `eos_lockfile_is_current` compared only name and version, so a changed `source`, `hash`, or `build_type` left a stale lockfile reported as current. Comparing the full tuple plus `project_version` is right.

I checked the bound that worried me — `int matched[EOS_MAX_PACKAGES]` indexed by `j < lock->count`. It is safe: `eos_lockfile_load` caps `lock->count` at `EOS_MAX_PACKAGES` when parsing. Worth a one-line comment saying so, since the safety depends on a different function.

Factoring the hash computation into `package_hash()` so generation and checking cannot drift apart is the right structural change — that class of bug (two copies of "how we compute the identity") is how lockfiles quietly stop working.

**Merge order:** this conflicts with #61 in `tests/test_config.c` — you both append a test at the same point. #61 fixes a segfault, so I'd take that first and rebase this on top. The resolution is mechanical.

srpatcha (approved):

The real defect is that `is_current` compared only name and version, so changing a package's `source`, `hash` or `build_type` left a stale lockfile reported as current. Comparing the full tuple fixes that, and hoisting `package_hash()` removes the duplicated derive-a-hash-from-metadata logic so generate and compare cannot drift apart — which is what would otherwise make every comparison fail for packages with no explicit hash.

Checked the new `matched[EOS_MAX_PACKAGES]`: `j` runs to `lock->count`, and `eos_lockfile_load` bounds that with `lock->count < EOS_MAX_PACKAGES` at line 118, so the index is in range. The flag also correctly stops one lock entry satisfying two identically-named config packages.

PR #47, opened 2026-08-26
docs(cmake): update build and test instructions with EOS_BUILD_TESTS

Diksha18Singh:
CLAUDE.md links to CMake test instructions, but the default configuration resulted in "No tests were found" without the explicit opt-in flag — confirmed via clean test runs.
This updates the build and test documentation, verified end-to-end from a clean checkout:

CMake test setup: added -DEOS_BUILD_TESTS=ON (default OFF) and updated the build directory to build/host

Clean build and test execution: verified all 21/21 host unit tests compiled and passed successfully (ctest --test-dir build/host --output-on-failure)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -136,7 +136,10 @@ read differently is a defect in the rule, and it will recur until it is fixed.
 ## CMake
 
 ```bash
-cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug
-cmake --build build -j
-ctest --test-dir build --output-on-failure
+cmake -B build/host -DCMAKE_BUILD_TYPE=Debug -DEOS_BUILD_TESTS=ON
+cmake --build build/host --parallel
+ctest --test-dir build/host --output-on-failure
 ```
+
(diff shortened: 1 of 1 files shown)


srpatcha:

The documented commands are correct — I ran them. `cmake -B build/host -DCMAKE_BUILD_TYPE=Debug -DEOS_BUILD_TESTS=ON` followed by `ctest --test-dir build/host` gives **24/24 pass**, and omitting `-DEOS_BUILD_TESTS=ON` does produce "No tests were found", exactly as your note says.

This is more valuable than its size suggests. Two CI jobs were passing `-DBUILD_TESTS=OFF` — the wrong option name — and it was silently ignored for the same reason your note documents. #68 fixes those.

srpatcha (approved):
Verified locally against current master: merges clean, cmake configure + ninja build with no new errors, `ctest` 24/24 passing. Docs-only change to the build/test instructions; the `EOS_BUILD_TESTS` flag it documents is the one the CMakeLists actually defines.

PR #49, opened 2026-08-26
fix(kernel): fix 32-bit tick counter overflow, add task runtime stats API

yash161:
## Issue

`eos_task_wake_check()` and `eos_swtimer_tick()` compared ticks with plain `current_tick >= deadline`. `g_tick` is a 32-bit counter, so at a 1kHz tick rate it wraps roughly every 49.7 days. Immediately after wraparound, `current_tick` (small, near 0) is numerically less than a still-pending `deadline` (large, near `UINT32_MAX`), so the comparison is false when it should be true — blocked tasks and software timers with deadlines spanning the wrap silently fail to fire (or, depending on which side of the wrap, fire early). This is a classic long-uptime bug that's easy to miss in short-running tests.

There was also no built-in way to inspect per-task runtime state (state, scheduled count, stack usage) at runtime, which makes debugging stack pressure or starvation on constrained targets harder than it needs to be.

## Approach

- Added `tick_expired(current, deadline)` in `kernel/src/task.c`, using the standard unsigned-subtraction technique (`(int32_t)(current - deadline) >= 0`) used by FreeRTOS, Zephyr, and the Linux kernel's `time_after()` — correct for deadline distances up to `INT32_MAX` ticks (~24.8 days at 1kHz), which comfortably covers realistic timer/sleep durations.
- Replaced the two direct `current_tick >= deadline` comparisons (in `eos_task_wake_check()` and `eos_swtimer_tick()`) with calls to `tick_expired()`.
- Added `eos_task_get_stats()` / `eos_task_get_all_stats()` and an `eos_task_stats_t` snapshot struct (id, name, priority, state, run_count, stack_used, stack_size) to `kernel/include/eos/kernel.h` and `kernel/src/task.c`. Stack usage is estimated by scanning the task's stack buffer from the base upward for the first non-zero word (tasks are allocated with zeroed stacks and a canary at word 0).

## Testing

- Added `test_tick_overflow()` to `tests/test_kernel.c`, which sets a task's `wake_tick` near `UINT32_MAX`, advances `current_tick` past the wrap, and asserts the task is woken correctly — this test fails against the old `>=` comparison and passes with `tick_expired()`.
- Added `test_task_stats()` covering both `eos_task_get_stats()` and `eos_task_get_all_stats()`, checking run_count increments and reported stack fields.
- Since CMake wasn't available in the review environment, compiled and linked `test_kernel.c` directly against the real kernel/HAL sources with `gcc` (no stubs): `kernel/src/{task,sync,ipc,multicore,mem/heap}.c` + `hal/src/{hal_common,hal_extended_stubs,hal_linux}.c`. Ran the resulting binary: **11/11 tests pass** (9 pre-existing + 2 new), including the overflow case.

## Limitations

- `tick_expired()` is correct for deadlines up to `INT32_MAX` ticks in the future; a caller scheduling something further out than that (unusual — ~24.8 days at 1kHz) would need a wider tick type, which is unchanged by this PR.
- Stack usage estimation in `eos_task_get_stats()` is a heuristic (scans for the first non-zero word from the stack base) — it can overestimate free stack if a task legitimately writes zero words near its high-water mark, consistent with how similar "high water mark" estimators work in other RTOSes.


srpatcha:
### Review — please split this into two PRs

Built merged onto current `master`: **24/24 ctest pass.** Both halves work; the request is about scope, not correctness.

**The tick fix.** `tick_expired()` is correct, and the comment citing FreeRTOS/Zephyr/`time_after` is the right reference. But it keeps `wake_tick > 0` as the "is a deadline set" test:

```c
if (t->state == EOS_TASK_BLOCKED && t->wake_tick > 0 &&
    tick_expired(current_tick, t->wake_tick)) {
```

so a deadline that lands exactly on tick 0 is still read as "wait forever" and the task never wakes. #63 separates the two meanings with a `wake_armed` flag and has a test that pins that case. If you keep this half, adopting that split would make it complete.

**The runtime stats API.** This is a genuinely useful feature and unrelated to a tick-overflow fix. It should be its own PR so it can be reviewed on its own merits — and because a `feat` mixed into a `fix` is hard to revert selectively.

One concrete defect in it, worth fixing before it lands anywhere:

```c
/* Skip word 0 (canary), count zero-filled words from bottom up */
for (uint32_t w = 1; w < total_words; w++) {
    if (g_tasks[h].stack_base[w] == 0) unused_words++;
    else break;
}
```

This assumes the stack region is zero-filled. `g_stack_pool` is `static`, so it is zeroed **once at load**. But `eos_kernel_init()` resets `g_stack_pool_offset = 0` (`kernel/src/task.c:198`) **without clearing the pool**, so after a re-init — or after task deletion and re-creation — a new task's stack contains the previous occupant's residue and `stack_used` reports a high-water mark that is not its own. Either zero the region in `alloc_task_stack`, or fill it with a known pattern and scan for that instead.

The zero-fill assumption also deserves a line in the doc comment: on a target where the stack region is not zeroed, this returns a meaningless number rather than an obviously wrong one.

Splitting is the ask. Both halves are worth merging.

srpatcha:
## Half of this is now redundant; the other half is worth keeping on its own

The tick-overflow part has landed via #63, which merged a few minutes ago. Three PRs found that bug independently (this one, #56, and #63), and #63 went furthest — besides the signed-difference comparison it also removed the `wake_tick > 0` sentinel, which was separately broken: a deadline landing exactly on tick 0 after a wrap read as "no deadline" and the task never woke. That is now an explicit `wake_armed` flag.

So this branch conflicts with master, and re-applying the tick half would be re-doing work.

**But the runtime statistics API in here is not a duplicate of anything**, and I would like it:

```c
typedef struct { ... uint32_t run_count; } eos_task_stats_t;
int eos_task_get_stats(eos_task_handle_t handle, eos_task_stats_t *out);
int eos_task_get_all_stats(eos_task_stats_t *out, int max_entries, int *count);
```

`run_count` is already maintained in `eos_task_t` and there is currently no way to read it — the field exists and nothing exposes it. An accessor pair is the natural thing to add, and `eos_task_get_all_stats` taking `max_entries` and writing back a `count` is the right shape for a caller with a fixed buffer.

Could you rebase onto current master and drop the tick changes, leaving just the stats API? That would be a clean, self-contained PR and I will merge it. Two things I would check while you are in there:

- `eos_task_get_stats` should reject an out-of-range handle and a NULL `out` the way the neighbouring accessors do.
- Worth deciding whether a `DELETED` task appears in `get_all_stats`. `eos_task_get_name` was just changed in #64 to report deleted handles as invalid, so consistency argues for skipping them here too.

Happy to look again as soon as it is up.

srpatcha (dismissed):

PR #50, opened 2026-08-26
fix(devicetree): bounds-check untrusted DTB input; enable the dead fuzz target

Kartikey1306:
## Summary

A flattened device tree is untrusted input. It arrives from a prior boot stage
or a flash region, and every offset and length inside the blob is chosen by
whoever produced it. `eos_dt_parse()` consumed those values without validating
them, so a malformed DTB reads outside the buffer in four distinct ways.

The fuzz harness that should have caught this was dead twice over — see
*Why this was not caught* below.

## The bugs

Reproduced with crafted blobs under `-fsanitize=address,undefined`:

| # | Malformed input | What happened | Site |
|---|---|---|---|
| 1 | `off_struct` > buffer length | loop bound `pos < (size - off_struct)` wraps → walk runs unbounded | `devicetree.c:37` |
| 2 | `BEGIN_NODE` name with no terminator | `strlen()` / `strncpy()` run past the allocation | `devicetree.c:48` |
| 3 | property `nameoff` = `0x40000000` | `strings + nameoff` dereferenced far outside the mapping | `devicetree.c:77` |
| 4 | nesting deeper than 32 | write guarded by `if (depth < 32)`, but `stack[depth - 1]` read unguarded | `devicetree.c:55` |

Case 4 is the one worth calling out — the array write *was* bounds-checked, so
it reads as safe on a skim. `depth` kept incrementing past the guard, and the
next `BEGIN_NODE` read `stack[32]` off the end of a stack array:

```
runtime error: index 32 out of bounds for type 'EosDtNode *[32]'
ERROR: AddressSanitizer: stack-buffer-overflow ... READ of size 8
    #0 eos_dt_parse devicetree.c:55
```

Two accessors had the same class of problem:

* `eos_dt_find_compatible()` called `strstr()` on `prop->data`. A property value
  is raw bytes; one that filled the 256-byte buffer has no terminator, so the
  search runs into whatever follows.
* `eos_dt_get_irq()` computed `index * 4` and used it as an offset *before*
  range-checking, so a negative index became a huge unsigned offset.

## Approach

Validate before dereferencing, everywhere:

* **Header offsets** — `totalsize` may not exceed the buffer actually passed in,
  and both blocks must begin inside the blob. The declared `size_dt_struct` /
  `size_dt_strings` are honoured only when they fit, otherwise clamped to the
  rest of the blob. Either way the walk stays in bounds.
* **Three small primitives** carry the checks so no call site can forget them:
  `dt_read_be32()` (4 bytes inside the block), `dt_strlen_at()` (terminator
  found before the block ends), `dt_align4()` (refuses the wrapping cases).
* **Depth** — nesting beyond `EOS_DT_MAX_DEPTH` is rejected rather than clamped.
  Silently dropping nodes would produce a tree that does not describe the blob.
* **Termination** — `pos` advances by ≥4 every iteration and every read is
  bounds-checked, so the loop terminates on any input. A blob with no `FDT_END`
  is rejected rather than walked to the end.
* **Property values are bytes, not strings.** `dt_prop_contains()` searches
  within `prop->len`. I kept substring semantics rather than switching to proper
  `compatible` string-list matching — that is a behaviour change, and this PR is
  about memory safety. Noted under *Limitations*.

`EOS_DT_MAX_DEPTH`, `EOS_DT_MAX_CHILDREN` and `EOS_DT_HEADER_SIZE` are now named
constants in the header instead of the literals `32`, `16` and `40`.

**The public API is unchanged.** Well-formed blobs parse exactly as before; the
only behavioural difference is that malformed ones now return `-1`.

## Why this was not caught

`tests/fuzz/fuzz_devicetree.c` exists and targets this exact parser, but:

1. **`tests/fuzz/` was never added by any `add_subdirectory`.** No fuzz target
   in the repo could be built — not `fuzz_devicetree`, not `fuzz_sha256`, not
   `fuzz_aes`, despite the CMakeLists advertising them.
2. `fuzz_devicetree` was *additionally* commented out, with the note "these
   require the parser libraries to be linked" — the `eos_devicetree` target it
   names does exist and links fine.
3. The harness declared `extern int eos_dtb_parse(const void *, size_t)`, which
   does not exist anywhere in the tree. The real entry point is
   `eos_dt_parse(EosDeviceTree *, const uint8_t *, uint32_t)`.

All three are fixed. `add_subdirectory(fuzz)` is wired in, the target is
uncommented, and the harness calls the real API.

## Testing

**New: `tests/test_devicetree.c`** — 37 assertions, wired into ctest.

* Happy path against a hand-built valid v17 blob: parse, node count, cached
  `model`/`compatible`, path lookup, parent links, and the `reg` / `u32` /
  `string` / `irq` / `phandle` accessors.
* One case per bug above, plus bad magic, short blob, `totalsize` larger than
  the buffer, unbalanced `END_NODE`, property length past the struct block, and
  a missing `FDT_END`.
* **Truncation sweep** — every prefix of a valid blob must be rejected, with the
  untruncated blob still parsing so the loop proves something.
* **Corruption sweep** — every byte replaced with each of 5 patterns
  (~2,000 malformed inputs); each must parse or be rejected, never over-read.

```console
$ cmake -S . -B build -DEOS_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug
$ cmake --build build -j                    # no new warnings under -Wall -Wextra
$ ctest --test-dir build --output-on-failure
100% tests passed out of 22                 # was 21
$ ./build/tests/test_devicetree
37/37 assertions passed
```

**Sanitizers — the same suite against the pre-fix parser:**

```console
$ clang -fsanitize=address,undefined -fno-sanitize-recover=all \
    tests/test_devicetree.c <parser> -o t

  with this change:  37/37 assertions passed          exit 0
  pre-fix parser:    ERROR: AddressSanitizer: heap-buffer-overflow
                     SUMMARY: ... in eos_dt_parse    exit 134 (SIGABRT)
```

**Mutation fuzzing** (libFuzzer is unavailable on this host — AppleClang ships
no `libclang_rt.fuzzer`, so I wrote a standalone ASan+UBSan mutation driver as a
stand-in). It mutates a valid blob, generates random bodies with correct magic,
and generates fully random buffers, then exercises the accessors on whatever
tree comes out:

```console
  pre-fix parser:   ERROR: AddressSanitizer: BUS ... in eos_dt_parse   (immediate)
  with this change: 5 seeds x 200-300k iterations = 1.1M inputs
                    ~9,400 accepted per run, zero sanitizer reports
```

`fuzz_devicetree` itself is **NOT RUN** here for the reason above; it is
verified to compile against the real API and the CMake wiring is verified to
configure (`-- Fuzz targets: fuzz_sha256, fuzz_aes, fuzz_devicetree`).

## Limitations and considerations

* **Stricter than before, deliberately.** Blobs with an unbalanced `END_NODE`,
  no `FDT_END`, or nesting past 32 are now rejected rather than partially
  parsed. All are malformed per the FDT spec; a partially-built tree is worse
  than a clear failure. If any board ships such a blob this would surface as a
  parse failure rather than silent corruption.
* **`eos_dt_find_compatible()` still does a substring match**, not proper
  `compatible` string-list matching. `strstr("arm,cortex")` matching inside
  `"arm,cortex-a53"` is arguably wrong, but changing it is a behaviour change
  beyond this fix. Worth a follow-up.
* **Node/property/child limits still silently truncate.** A node with more than
  16 properties keeps the first 16, as before. I did not change this — it is a
  capacity policy, not a safety issue, and changing it would alter behaviour for
  valid blobs.
* `eos_dt_parse()` has no callers in-tree today; it is public API. The fix
  matters for whoever wires it into the boot path.
* Sanitizers are not part of the default CMake options here, so CI runs the new
  suite unsanitized. It still catches all four bugs (the pre-fix parser aborts),
  but adding an `EOS_SANITIZE` option would be a reasonable follow-up.

## Type of Change

- [x] fix — Bug fix
- [x] test — Add or fix tests
- [x] build — Build system changes
- [x] docs — Documentation

## Changes

- `drivers/devicetree/devicetree.c` — bounds-checked parse; hardened
  `eos_dt_find_compatible()` and `eos_dt_get_irq()`.
- `drivers/devicetree/devicetree.h` — `EOS_DT_MAX_DEPTH`,
  `EOS_DT_MAX_CHILDREN`, `EOS_DT_HEADER_SIZE`.
- `tests/test_devicetree.c` — new suite (37 assertions).
- `tests/CMakeLists.txt` — register `test_devicetree`; `add_subdirectory(fuzz)`.
- `tests/fuzz/CMakeLists.txt` — enable `fuzz_devicetree`.
- `tests/fuzz/fuzz_devicetree.c` — call the real API.
- `docs/security-hardening-guide.md` — section 13, *Parsing Untrusted Binary Input*.
- `CHANGELOG.md` — `Unreleased` entries.

## Pre-Submission Checklist

- [x] Compiles without warnings (`-Wall -Wextra`)
- [x] All existing tests pass (22/22, was 21)
- [x] New tests added, verified to abort on the pre-fix parser
- [x] Documentation updated
- [x] Public API unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 - **`eos_queue_create`:** Size check now uses division so `item_size * capacity` cannot wrap `size_t` on 32-bit targets and overflow the 1024-byte queue store.
 - **`eos_sem_create`:** Reject `initial > max` and `max` values that do not fit in `int32_t`, so the counting-semaphore invariant cannot be created already broken.
 - **`eos_mutex_lock`:** Recursive lock returns `EOS_KERN_FULL` at `uint8_t` saturation instead of wrapping `rec_count` to 0 and leaving the mutex stuck.
+- **`eos_dt_parse`:** Bounds-check the flattened device tree blob before dereferencing it. A malformed DTB could previously read outside the buffer four different ways: `off_struct` past the end of the blob (the `size - off_struct` bound wrapped), a node name with no terminator (unbounded `strlen`), a property name offset outside the strings block, and nesting deeper than the 32-entry node stack. All are now rejected.
+- **`eos_dt_find_compatible`:** Search within `prop->len` instead of calling `strstr()` on a property value that is not guaranteed to be NUL-terminated.
+- **`eos_dt_get_irq`:** Reject a negative index, and one large enough that `index * 4` overflows, before it is used as an offset.
+
+### Added
+- `tests/test_devicetree.c` — device tree parser suite: happy path, malformed-blob rejection, a sweep over every truncation of a valid blob, and a single-byte corruption sweep.
+- `tests/fuzz/` is now wired into the build. The directory was never added by any `add_subdirectory`, so no fuzz target could be built. `fuzz_devicetree` is enabled and calls the real `eos_dt_parse()` API — it previously declared a non-existent `eos_dtb_parse()` and was commented out.
 
 ## [3.0.1] - 2026-05-16
 
--- a/docs/security-hardening-guide.md
+++ b/docs/security-hardening-guide.md
@@ -255,6 +255,41 @@ if (remaining < EOS_STACK_CRITICAL_THRESHOLD) {
 
 ---
 
+## 13. Parsing Untrusted Binary Input
+
+A flattened device tree is not trusted input. It arrives from a prior boot
+stage or a flash region, and every offset and length inside the blob is chosen
+by whoever produced it. `eos_dt_parse()` therefore validates before it
+dereferences:
+
+- `totalsize` may not exceed the buffer length passed in, and the struct and
+  strings blocks must both begin inside the blob. Without this the
+  `size - off_struct` arithmetic wraps and the walk runs off the end.
+- Every token, property length and property name offset is bounds-checked
+  against its block before it is read.
+- Node names and property names must be NUL-terminated inside the blob;
+  an unterminated name is rejected rather than passed to `strlen()`.
+- Nesting deeper than `EOS_DT_MAX_DEPTH` is rejected rather than walking past
+  the parser's own node stack.
+- Property values are raw bytes and are not assumed to be NUL-terminated.
+  `eos_dt_find_compatible()` searches within `prop->len` instead of calling
+  `strstr()` on the buffer.
+
+Regression coverage lives in `tests/test_devicetree.c`, which sweeps every
+truncation of a valid blob and every single-byte corruption of it. A libFuzzer
+target is available for deeper coverage:
+
+```bash
+cmake -B build -DEOS_BUILD_TESTS=ON -DEOS_ENABLE_FUZZING=ON \
+      -DCMAKE_C_COMPILER=clang
+cmake --build build --target fuzz_devicetree
+./build/tests/fuzz/fuzz_devicetree -max_total_time=60
+```
+
+Apply the same rules to any new parser that consumes off-device data: validate
+offsets against the buffer you were actually handed, never trust a length field,
+and never assume a byte range is NUL-terminated.
+
 ## Production Build Checklist
 
 Use this checklist before releasing a production firmware image:
@@ -274,3 +309,4 @@ Use this checklist before releasing a production firmware image:
 - [ ] `EOS_LOG_LEVEL` is `EOS_LOG_ERROR` or `EOS_LOG_NONE`
 - [ ] Rollback protection counter is set correctly
 - [ ] SBOM is generated and archived for this build
+- [ ] Parsers for off-device input are fuzzed (see section 13)
--- a/drivers/devicetree/devicetree.c
+++ b/drivers/devicetree/devicetree.c
@@ -11,99 +11,212 @@ static uint32_t be32(const uint8_t *p) {
            ((uint32_t)p[2] << 8)  | (uint32_t)p[3];
 }
 
-static uint32_t align4(uint32_t v) { return (v + 3) & ~3u; }
+/* ------------------------------------------------------------------
+ * Bounds-checked primitives
+ *
+ * Everything below treats the blob as untrusted. A flattened device tree
+ * usually arrives from a prior boot stage or a flash region, so every offset
+ * and length it contains has to be validated against the buffer we were
+ * actually handed before it is used to index into it.
+ * ------------------------------------------------------------------ */
 
-int eos_dt_parse(EosDeviceTree *dt, const uint8_t *dtb, uint32_t size) {
-    if (!dt || !dtb || size < 40) return -1;
+/* Round `v` up to a 4-byte boundary, refusing the cases where that wraps. */
+static int dt_align4(uint32_t v, uint32_t *out)
+{
+    if (v > UINT32_MAX - 3u) return -1;
+    *out = (v + 3u) & ~3u;
+    return 0;
+}
+
+/* Read a big-endian u32 at `off`, provided all four bytes are inside `len`. */
+static int dt_read_be32(const uint8_t *buf, uint32_t len, uint32_t off,
+                        uint32_t *out)
+{
+    if (off > len || len - off < 4u) return -1;
+    *out = be32(buf + off);
+    return 0;
+}
+
+/* Length of the NUL-terminated string at `off`, without reading past `len`.
+ * Returns -1 when the buffer ends before a terminator is found. */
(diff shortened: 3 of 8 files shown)


srpatcha:

Built merged onto current `master`: **25/25 ctest pass** (your new devicetree test brings it from 24).

This is thorough work. The three things I'd single out:

1. **`dt_align4` refusing to wrap** rather than returning a wrapped value. Most implementations of this get it wrong.
2. **Not assuming property values are NUL-terminated** and searching within `prop->len` in `eos_dt_find_compatible` instead of `strstr`. That is the bug class that bites people in production.
3. **Wiring up `tests/fuzz/`.** No `add_subdirectory` had ever included it, so four fuzz targets had never been built. That fix is worth merging independently of the parser work.

The truncation sweep and single-byte-corruption sweep are the right shape for a parser test.

**One thing to know before you invest more here:** ADR-013 in #69 proposes replacing the hand-written FDT walker with `libfdt`, on the grounds that this PR is re-deriving bounds checks that a standard library has had for years. **That is not a reason to close this** — your test corpus and the fuzz wiring stay valuable under libfdt and become its acceptance test, and this fixes a real vulnerability that exists today while the ADR is still Proposed. Merge this; the ADR is a separate decision on a longer timescale.

Kartikey1306:
Hi EmbeddedOS Team,

Thank you for reviewing the eOS PR and for the detailed feedback. I’m glad to see that the updated test suite is passing successfully and that the contribution is ready to be merged.

Since I’ve now completed the assessment and the submitted PRs have been reviewed and accepted, I wanted to understand more about the role itself and the next stage of the process.

Could you please share some more details about the position, particularly the key responsibilities, technical requirements, expected level of ownership, working arrangement, and what you’re looking for in the person joining the team?

I’d also appreciate knowing whether, based on my assessment contributions and background, my profile aligns with the requirements for the role and whether I would be considered for the next stage/interview.

I’m happy to provide any additional information or discuss my contributions and experience in a call.

Best regards,
Kartikeydheer Srivastava

srpatcha (approved):
Verified locally against current master: merges clean, builds with no new errors, `ctest` 25/25.

A DTB is off-device input, so every one of these is reachable by whoever supplies the blob. The four read paths are all real and independent, and the one I would have most expected to be missed is `size - off_struct` wrapping when `off_struct` is past the end — an unsigned subtraction used as a bound is the classic way a length check becomes a no-op.

Two things I particularly want in:

- The overflow guards on the arithmetic itself (`v > UINT32_MAX - 3u` before aligning, `index * 4` in `eos_dt_get_irq`) rather than only on the resulting offsets. Checking the product after computing it is checking a value that already wrapped.
- The note that `pos` advances by at least 4 every iteration. Termination is the other half of bounds-safety in a token loop, and stating the invariant is what lets a reader confirm it rather than assume it.

`tests/fuzz/` never being reachable by any `add_subdirectory`, and `fuzz_devicetree` declaring an `eos_dtb_parse()` that does not exist, is worth calling out on its own — the repo had a fuzzing story that could not have compiled. Wiring it to the real `eos_dt_parse()` turns it from documentation into a check.

Not verified here: the libFuzzer run. No fuzzing build on this machine, so I have the unit tests and the review, not fuzz coverage.

Kartikey1306:
Hi EmbeddedOS Team,

Thank you for the detailed review and for independently verifying the changes against the current master. I appreciate the additional validation of the overflow handling, parser termination checks, and the fuzzing integration. I’m glad to see the changes are cleanly building and that 25/25 tests are passing.

With the technical assessment now completed and the contributions reviewed and approved, I wanted to understand more clearly what the next steps are for the role.

Could you please share some details about the position, including:

1. The role/title and primary responsibilities
2. The main technical areas and projects I would be working on
3. The expected level of ownership and contribution
4. Whether the position is full-time, part-time, or contract
5. The expected working hours/time-zone and collaboration requirements
6. The compensation range or structure
7. The remaining interview/selection stages

Based on the work I’ve completed across eBuild, eOS, and eBoot, including the five accepted PRs and the testing and security work involved, I’d also appreciate your view on whether my profile aligns with what you’re looking for and whether I can proceed to the next stage.

I’m available to discuss the assessment, walk through my implementation decisions, or complete any further technical interview or evaluation required.

Thank you, and I look forward to hearing from you regarding the role and next steps.

Best regards,
Kartikeydheer Srivastava
GitHub: Kartikey1306

PR #52, opened 2026-08-26
Fix package test stack overflow

Sougata0100:
## Summary

Fixes a stack overflow issue in the package tests.

### Changes

* Changed `EosPackageSet set` from automatic (stack) storage to `static` storage in the affected package test functions.
* This prevents the large `EosPackageSet` object from consuming excessive stack space during test execution.
* No production eOS code was modified.

### Testing

* Ran the complete CTest suite.
* **21/21 tests passed (100%)**.
* Total test time: **1.32 seconds**.

### Commit

`838b759` — `Fix package test stack overflow`

### DIFF
--- a/tests/test_package.c
+++ b/tests/test_package.c
@@ -10,7 +10,7 @@ static int passed = 0;
 static int dummy_init(void) { return 0; }
 
 static void test_package_register(void) {
-    EosPackageSet set;
+    static EosPackageSet set;
     memset(&set, 0, sizeof(set));
     EosPackage pkg;
     memset(&pkg, 0, sizeof(pkg));
@@ -24,7 +24,7 @@ static void test_package_register(void) {
 }
 
 static void test_package_find(void) {
-    EosPackageSet set;
+    static EosPackageSet set;
     memset(&set, 0, sizeof(set));
     EosPackage pkg;
     memset(&pkg, 0, sizeof(pkg));
@@ -39,7 +39,7 @@ static void test_package_find(void) {
 }
 
 static void test_package_find_not_found(void) {
-    EosPackageSet set;
+    static EosPackageSet set;
     memset(&set, 0, sizeof(set));
     const EosPackage *found = eos_package_find(&set, "nonexistent");
     assert(found == NULL);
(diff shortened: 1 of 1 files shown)


srpatcha:
### Review — verified correct; overlaps with #48

Built merged onto current `master`: **24/24 ctest pass.** The fix is right: `sizeof(EosPackageSet)` is ~3.9 MB and overflows the default 1 MB thread stack on Windows and other small-stack hosts.

Being straight with you: this change is **byte-identical** to the same hunk in #48 (I compared git blob hashes — both produce `de4400a`). #48 additionally carries a comment explaining *why* the sets are `static`, which is what stops someone reverting it later, so that is the one I'd suggest merging.

**That does not make this PR wrong or unwelcome** — you found a real defect and fixed it correctly, and you got there independently. Two options, both fine:

- Leave this open; once #48 merges, this rebases to an empty diff and can close itself out naturally.
- Or, if you'd like this to be the one that lands, add the explanatory comment and say so — the maintainers can take yours instead.

If you want a follow-up with more surface: `tests/test_heap.c` and several other test files have related issues, and 13 test files in `tests/` are not registered in any `CMakeLists.txt` at all, so they have never been compiled or run.

srpatcha:
## Same fix as #48, which just merged — yours was correct

You and #48 independently made the identical change: every `EosPackageSet` in `tests/test_package.c` moved from automatic to `static`. I verified both against master and both were green at 24/24.

I merged #48 for one reason, and it is not the code — the two diffs are byte-for-byte the same on every line that matters. #48 carries a comment explaining *why*:

> `EosPackageSet` is `EOS_MAX_PACKAGES * sizeof(EosPackage)` = ~3.8 MB, which overflows the 1 MB default thread stack on Windows and on any other small-stack host, crashing these tests before `main()` gets to run them.

That is the part a future reader needs. A bare `static` on a test-local invites someone to "tidy" it back to an automatic and silently reintroduce the crash, and it also raises the obvious question of whether state now leaks between tests — which the same comment answers by pointing at the `memset` on entry.

Nothing wrong with this PR; it is a coin toss decided by documentation. The diagnosis was right and the fix was right.

If you would like a follow-up in the same area: `cmd/eos/main.c` declares the same type the same way, and there is no assertion anywhere that `EosPackageSet` stays within a sane size. A static assert on `sizeof(EosPackageSet)` would turn the next growth of `EOS_MAX_PACKAGES` into a compile error rather than a stack crash on one platform. Happy to review that if you want to pick it up.

srpatcha (approved):

srpatcha (approved):

PR #53, opened 2026-08-26
Validate network socket protocol

akashdumala:
## Summary

Validate the protocol passed to `eos_net_socket()`.

## Changes

- Return `EOS_SOCKET_INVALID` for unsupported protocols.
- Added a regression test for an invalid protocol value.

## Testing

- All 16 network tests passed.
- `git diff --check` passes cleanly.

### DIFF
--- a/net/src/net.c
+++ b/net/src/net.c
@@ -27,8 +27,14 @@ void eos_net_deinit(void)
 
 eos_socket_t eos_net_socket(eos_net_proto_t proto)
 {
-    (void)proto;
-    if (!net_initialized) return EOS_SOCKET_INVALID;
+    if (!net_initialized) {
+        return EOS_SOCKET_INVALID;
+    }
+
+    if (proto != EOS_NET_TCP && proto != EOS_NET_UDP) {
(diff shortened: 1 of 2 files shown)


srpatcha:

Correct and minimal. Validating the protocol argument before allocating a descriptor is right.

Context you may want, not a criticism of this PR: `eos_net_connect()` currently returns `-1` unconditionally and `eos_net_socket()` returns an incrementing integer — the subsystem is an API surface with no protocol implementation behind it. Your validation is still correct and worth merging; I mention it only so the passing test isn't read as "networking works".

Non-blocking: `static int next_fd` never resets on `eos_net_deinit()`, so descriptors keep climbing across init/deinit cycles. Pre-existing, and a reasonable follow-up if you want it.

srpatcha (approved):

`proto` was explicitly discarded with `(void)proto`, so any value — including one out of the enum's range — allocated a descriptor. Validating it before handing back an fd is right, and it also removes the misleading `(void)` marker that said the parameter was intentionally unused. Test covers the out-of-range case rather than only the two valid ones.

PR #55, opened 2026-08-26
fix(kernel): restore mutex owner priority after waiter timeout

Noor-Nabi4:
Priority inheritance boosted the owner when a higher-priority task
blocked, but a timeout left that boost in place and could starve
other work. Recompute owner priority from remaining waiters, and
reject a full waiter table without applying a boost.

Signed-off-by: Noor Nabi <gcnoor6@gmail.com>
Co-authored-by: Cursor <cursoragent@cursor.com>


srpatcha:
### Review — correct fix, verified; overlaps with #67

The bug is real and well diagnosed: a waiter that times out was removed from the queue but the owner kept the inherited boost, so a low-priority task could run elevated indefinitely after the waiter that justified it had gone.

Two things you did that I'd highlight:

- **Refusing the lock with `EOS_KERN_NO_MEMORY` when the waiter table is full**, instead of boosting and then blocking a caller that was never enqueued. That was a lost-wakeup waiting to happen with `EOS_WAIT_FOREVER`.
- **The `eos_port_yield` hook that samples owner priority mid-lock.** With a no-op yield the boosted state exists only during that call, so observing it from inside the stub is the only way to test this. Neat.

**Overlap:** #67 rebuilds priority inheritance around a recomputed invariant (`effective = min(base, all waiters on all held mutexes)`) and adds transitive propagation. It covers this timeout case and the nested ones. I verified the two conflict in `kernel/src/sync.c`.

**Not a reason to close this.** Two paths, both reasonable:

- If #67 lands first, rebase this and keep anything it misses. Specifically: `mtx_recompute_owner_priority` restoring from `m->original_prio` handles single-mutex ownership correctly, and your test is more focused than #67's suite — worth keeping as a targeted regression test either way.
- If the maintainers prefer the smaller change first, this is the lower-risk fix for the specific reported bug and can merge on its own.

One thing worth raising either way: `eos_task_set_current_internal` is exposed in `kernel_internal.h` purely so host tests can drive multi-task paths. That is legitimate, but it deserves the comment you gave it — it is a test seam in a shipped header.

srpatcha (dismissed):

srpatcha (approved):

PR #56, opened 2026-08-26
fix: resolve tick wrap-around bugs in scheduler and timers

saurabh850:
### The issue, bug, or improvement being addressed
The `g_tick` SysTick counter is a 32-bit integer that overflows roughly every 49.7 days when running at a 1ms tick rate. The existing code performed direct `>=` comparisons (`current_tick >= t->wake_tick` and `current_tick >= t->next_tick`) for kernel delays and software timers. Due to integer overflow, these conditions evaluated to true prematurely if a tick wrapped around during the wait period. Additionally, if the delay calculation exactly wrapped around to 0, the task was inadvertently suspended indefinitely because `wake_tick = 0` was utilized as a sentinel value for `EOS_WAIT_FOREVER`.

### Proposed approach and implementation
Replaced the direct comparisons with signed integer difference checks (`(int32_t)(current_tick - t->wake_tick) >= 0`) in `eos_task_wake_check` and `eos_swtimer_tick` to safely handle the 32-bit wrap-around. In `eos_task_delay_ms` and `eos_task_block_with_timeout`, added a check to ensure `wake_tick` never wraps identically to 0. If `wake_tick` calculates to 0, it is forced to 1, preserving the `EOS_WAIT_FOREVER` sentinel behavior without halting the task indefinitely.

### Testing or validation performed
Ran the automated test suite via `python run_all_tests.py`, which passed all 10 unit, functional, and performance tests, confirming that no regressions were introduced to the RTOS kernel pipeline and task scheduling mechanisms.

### Additional considerations or limitations
None at this time.

### DIFF
--- a/kernel/include/eos/kernel.h
+++ b/kernel/include/eos/kernel.h
@@ -86,6 +86,7 @@ typedef struct {
     uint32_t *stack_base;
     uint32_t *stack_ptr;
     uint32_t wake_tick;
+    bool wake_armed;
     uint32_t run_count;
 } eos_task_t;
 
--- a/kernel/include/eos/kernel_internal.h
+++ b/kernel/include/eos/kernel_internal.h
@@ -37,12 +37,12 @@ void eos_task_set_priority_internal(eos_task_handle_t h, uint8_t prio);
  * @brief Block a task with a timeout.
  *
  * Sets the task state to BLOCKED and wake_tick to g_tick + timeout_ms.
- * If timeout_ms is EOS_WAIT_FOREVER, wake_tick is set to 0 (never auto-wake).
+ * If timeout_ms is EOS_WAIT_FOREVER, wake_armed is set to false (never auto-wake).
  */
 void eos_task_block_with_timeout(eos_task_handle_t h, uint32_t timeout_ms);
 
 /**
- * @brief Unblock a task (set state to READY, clear wake_tick).
+ * @brief Unblock a task (set state to READY, clear wake_armed).
  */
 void eos_task_unblock(eos_task_handle_t h);
 
--- a/kernel/src/task.c
+++ b/kernel/src/task.c
@@ -153,7 +153,7 @@ void eos_task_wake_check(uint32_t current_tick)
     for (int i = 0; i < EOS_MAX_TASKS; i++) {
         eos_task_t *t = &g_tasks[i];
         if (t->state == EOS_TASK_BLOCKED && t->wake_tick > 0 &&
-            current_tick >= t->wake_tick) {
+            (int32_t)(current_tick - t->wake_tick) >= 0) {
             t->state = EOS_TASK_READY;
             t->wake_tick = 0;
         }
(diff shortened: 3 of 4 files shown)


srpatcha:
### Review — correct comparison fix; one case still open

The core change is right, and it is the standard technique: `(int32_t)(now - deadline) >= 0` instead of `now >= deadline`. Both call sites — `eos_task_wake_check` and `eos_swtimer_tick` — are the correct ones.

**The case that remains:** `wake_tick == 0` still doubles as the "wait forever" sentinel. Your patch handles the collision by bumping a deadline that lands on 0 to 1:

```c
if (g_tasks[g_current].wake_tick == 0) g_tasks[g_current].wake_tick = 1;
```

That works, and it costs at most one tick of accuracy. It is a fair engineering trade. But it leaves the two meanings of `wake_tick` entangled, so the next person to touch this has to rediscover why the `== 0` line exists.

#63 takes the other approach — a separate `wake_armed` flag, so "do I have a deadline" and "what is the deadline" stop sharing one field — and includes a test (`test_wake_deadline_of_zero`) that pins the case directly. I verified that test fails on unfixed `master`.

**Suggestion, and please don't read it as a rejection:** if #63 lands first, this rebases to almost nothing, and that is fine — you diagnosed the same real bug. If you'd like this to be the version that merges instead, adopting the `wake_armed` split here would make it strictly better than #63, because your patch also covers `eos_task_delay_ms`, which is a path #63 sets but does not test. Either way the diagnosis was right.

saurabh850:
Agreed on the trade-off. Splitting the state into a dedicated wake_armed flag makes a lot of sense.

I have updated this PR to adopt the wake_armed approach. This removes the wake_tick = 1 workaround and applies the decoupled logic across all paths, including eos_task_delay_ms.

srpatcha:
## The bug you found is fixed — this landed via #63

Your diagnosis was right, and three separate PRs (#63, #49, and this one) independently found the same wraparound. #63 has now merged, so `master` carries the fix:

```c
#define EOS_TICK_REACHED(now, deadline) \
    ((int32_t)((uint32_t)(now) - (uint32_t)(deadline)) >= 0)
```

applied in both `eos_task_wake_check()` and `eos_swtimer_tick()` — the scheduler *and* the timers, which is the same pair this PR covers. I checked `origin/master`'s `eos_swtimer_tick` specifically to be sure the timer half was not left behind:

```c
if (EOS_TICK_REACHED(current_tick, t->next_tick)) {
```

#63 also went one step further: `wake_tick > 0` was being used as the "is a deadline set" sentinel, so a deadline landing exactly on tick 0 after a wrap read as "no deadline" and the task never woke at all. That is replaced with an explicit `wake_armed` flag.

This branch now conflicts with master, and re-applying it would only re-do work that is already in. I would rather not leave it open as a phantom — but it is your call, so I am not closing it. If you would like something from it kept, say which part and I will make sure it lands; otherwise closing it as superseded by #63 is the tidy outcome.

Thanks for catching it — three independent reports of the same bug is a good signal that it was worth fixing properly.

saurabh850:
Makes total sense! Since #63 is merged and covers the core wraparound and state decoupling cleanly, there's no need to keep this open or re-do the work. I'll go ahead and close this out as superseded.

Thanks for the great architectural feedback on the wake_armed flag, really enjoyed digging into the scheduler on this one!

PR #57, opened 2026-08-26
feat: transition package verification from stubs to production-grade crypto

### LINKED ISSUE #17: feat: transition package verification from stubs to production-grade crypto
### Current Situation

I noticed during my review of the `eos_pkg` implementation that the package verification logic currently relies on "stub" functions. Specifically:

*   `eos_ed25519_verify_stub`: Always returns `0` (Success) without performing any cryptographic verification.
*   `eos_sha256_stub`: Implements a simple XOR/addition-based hashing algorithm instead of a cryptographically secure hash like SHA256.

### Problem Description

The use of these stubs in a project aiming for production-level embedded safety and security (as indicated by the architecture and products like `aerospace.h` and `medical.h`) creates a false sense of security. An attacker could easily craft a malicious `.eapp` package that satisfies the weak hash requirement and bypasses signature verification entirely.

### Proposed Improvement

I recommend prioritizing a transition from these stubs to production-grade cryptographic implementations:

1.  **Integrate a real Ed25519 library**: Replace the stub with a proper verification routine.
2.  **Use standard SHA256**: Replace the custom XOR-sum with the project's existing SHA256 implementation (which I saw in `services/crypto/`).
3.  **Harden the .eapp header**: Ensure the header contains all necessary metadata for full integrity and authenticity checks.

### Why This Matters

For a next-gen embedded OS, the integrity of the package management system is a foundational security property. Moving away from these placeholders will bring the project closer to its stated security goals and provide real protection for end-users.

---

I'm happy to discuss the architectural implications of these changes if there's interest in moving this forward.

RinZ27:
## Summary

This pull request integrates the `orlp/ed25519` library to perform proper signature verification and replaces the weak `eos_sha256_stub` with the existing SHA-256 implementation from `services/crypto/`. This directly addresses the vulnerability where malicious packages could bypass the weak stubs.

## Type of Change

- [ ] feat — New feature
- [x] fix — Bug fix
- [ ] test — Add or fix tests

## Changes

- Removed `eos_sha256_stub` and `eos_ed25519_verify_stub`.
- Integrated the `ed25519` library into `services/crypto/CMakeLists.txt`.
- Replaced stub calls with proper cryptographic routines in `services/pkg/eos_pkg.c`.
- Fixed duplicate syntax errors in `eos_pkg.h`.

## Testing

- [x] Unit tests pass (`ctest --test-dir build --output-on-failure`)
- [ ] Integration tests pass

## Pre-Submission Checklist

- [x] All existing tests pass
- [x] Branch is rebased on latest master

## Related Issues

Fixes #17

## Screenshots / Logs

N/A

## Additional Notes

I noticed the duplicate enums in `eos_pkg.h` causing syntax errors during compilation, so I took the liberty of cleaning those up as well.


srpatcha:
### Review — right problem, and the fix is blocked on provenance rather than code

**You found something real, and it is more serious than the PR title suggests.** An audit recorded in ADR-011 reached the same conclusion independently: `eBoot/core/ed25519_verify.c` is **not RFC 8032**. It computes the challenge as `SHA-256(R‖A‖M)` zero-padded to 64 bytes where the standard requires SHA-512, so both RFC 8032 §7.1 test vectors are **rejected**, and a signature produced by the project's own `tools/sign_image.py` — which uses real Ed25519 — is rejected too. Ed25519 secure boot has never worked end to end. It fails closed, so nothing forged is accepted, but the feature does not function.

So the instinct behind this PR is correct and the gap is worth closing.

**What blocks it as written:** the 4,815 added lines are third-party cryptographic code with **no licence header, no attribution, no upstream URL and no pinned revision**. I searched the whole diff for `copyright`, `licen`, and `public domain` — zero matches. For an MIT project that ships a `NOTICE` file, that is a compliance problem before it is a technical one. It also adds **no test vectors** for the verification path it introduces, which for a signature scheme is the one thing that must be there.

This is **not a criticism of you.** The repository had no documented way to bring external code in — no `third_party/` convention, no import procedure, nothing in `CONTRIBUTING.md`. Given that, pasting the source in was a reasonable thing to try. #70 adds exactly that missing procedure.

**Three ways forward, all of which keep your work:**

1. **Re-land under `third_party/` once #70 merges.** Add `UPSTREAM` (repo URL + full 40-char commit SHA + import date), copy `LICENSE` verbatim, add the entry to `NOTICE`, and add the RFC 8032 §7.1 vectors as a test. `scripts/check_third_party.py` in #70 will tell you exactly what is missing.
2. **Point at the upstream instead of vendoring** — CMake `FetchContent` pinned to a SHA. Smaller diff, same result, and CVE tracking comes with it.
3. **ADR-012 in #69** proposes Mbed TLS via the PSA Crypto API as the org-wide provider, partly because PSA maps onto hardware crypto accelerators on STM32/nRF/ESP32. If that is accepted, the eBoot verify path takes a reviewed subset rather than a vendored copy — and your finding is what forces that decision to actually happen.

Whichever route, please add the RFC 8032 vectors. ADR-011 item 4 is explicit: no crypto feature claims "Implemented" without published conformance vectors passing in CI. Right now `tests/unit/test_ed25519.c` passes while every assertion in it checks *rejection* of malformed input — negative-only tests are how a non-functional implementation survived this long.

srpatcha:
## Right direction, and I want this — but not without provenance and vectors

Verified locally against current master: merges clean, builds, `ctest` 26/26. Moving package verification off stubs onto real Ed25519 is the correct goal, and it matters more than usual here: eBoot currently ships an Ed25519 whose own tests document *"three measurably wrong primitives (point decompression, scalarmult-base, scalar reduction)"* and refuse to verify anything. Getting a correct implementation into the tree is genuinely valuable.

Four things block it as it stands, and none are about the code being wrong.

### 1. 4,815 lines of vendored crypto with no attribution

```
$ head -3 services/crypto/src/ed25519_fe.c
#include "fixedint.h"
#include "fe.h"
```

No copyright line, no licence, no SPDX identifier, no upstream named — in a repo where every other file starts with:

```c
// SPDX-License-Identifier: MIT
// Copyright (c) 2026 EoS Project
```

The file set (`fe.h`, `ge.h`, `sc.h`, `precomp_data.h`, `fixedint.h`, `load_3()`) looks like **orlp/ed25519**, itself derived from ref10/SUPERCOP. If that is right the licence is permissive and there is no problem to solve beyond recording it — but it has to be recorded, and I should not have to infer the origin of a signature verifier from its helper names.

### 2. NOTICE and LICENSE are untouched

Which means the repo cannot produce an accurate SBOM the moment this lands. That is not hypothetical: **#70** adds exactly that SBOM generation and a third-party import policy, and **#69** records the ADRs governing third-party adoption. This PR is the first real test of both, and it should go in *after* them, following the policy rather than ahead of it.

### 3. No RFC 8032 test vectors — no tests at all

```
$ git diff origin/master...pr57 --name-only | grep -i test
(nothing)
```

For a signature verifier this is the gap that worries me most. `ctest` being 26/26 means the existing suite still passes; it does not mean this code verifies signatures correctly, because nothing exercises it. A verifier that wrongly *rejects* is a broken update path; one that wrongly *accepts* is a security hole, and neither shows up in a build.

eBoot is the cautionary example — its broken Ed25519 was only pinned down because someone wrote `test_ed25519_rfc8032` against real vectors. Please add the RFC 8032 §7.1 vectors, including the edge cases: small-order points, non-canonical encodings, and `s` values above the group order.

### 4. Two implementations, one tree

After this, eos and eBoot each carry their own Ed25519 and one of them is known broken. Worth stating in the PR which is intended to survive, even if unifying them is a later change.

### What I would like

Land **#69** and **#70** first, then rebase this on top and let the policy tell you what the NOTICE entry and SBOM record should look like. Add the vectors. Then I will merge it.

Happy to review the vector tests separately if that is easier — that is the piece I most want to see, independent of the paperwork.

srpatcha (approved):

PR #59, opened 2026-08-26
fix(crypto): reject invalid AES key sizes and guard null crypto contexts

Divyanshu-2907:
## Summary

`eos_aes_init()` accepted any `int` for `key_bits` and never validated it. `EosAesCtx.rk` is a
60-word array — exactly what an AES-256 schedule needs — but the key-expansion loop is bounded by
`4 * (nr + 1)`, where `nr` derives from that unvalidated parameter. Nothing connects the array
bound to the value driving the loop, so an out-of-range key size runs the expansion off the end of
the caller's context. Three distinct memory-safety faults reproduce under
`-fsanitize=address,undefined`:

| `key_bits` | `nk` | Result |
|---|---|---|
| 512 | 16 | stack-buffer-overflow — 92 words written into `rk[60]`, 4-byte **WRITE** past the context |
| 0 | 0 | stack-buffer-underflow — reads `rk[-1]`, then evaluates `i % 0` |
| 64 | 2 | global-buffer-overflow — `rcon[i/nk]` reads past the 11-entry `rcon[]` table |

Separately, `eos_aes_*` and the four `eos_sha512_*` entry points dereference their context pointer
with no NULL check. `services/crypto/src/sha256.c` — same library, same header — already guards all
four of its public functions (`sha256.c:62,72,97,121`). Its siblings did not.

Two gaps kept this invisible: the test suites that would have caught it were never registered in
`tests/CMakeLists.txt`, and `ci.yml` — the only workflow that runs `ctest` — targets branches that
do not exist in this repository.

## Type of Change

- [x] `fix` — Bug fix
- [x] `test` — Add or fix tests
- [x] `ci` — CI/CD pipeline changes

## Changes

- **`services/crypto/src/aes.c`** — validate `key_bits` against `{128, 192, 256}` and reject a
  NULL key. A rejected context is zeroed, leaving `nr == 0` — never a valid round count — and
  `eos_aes_encrypt_block`, `eos_aes_decrypt_block`, `eos_aes_cbc_encrypt` and
  `eos_aes_cbc_decrypt` now refuse that sentinel. The intent is to **fail closed**: a caller that
  ignores a bad key size gets no ciphertext at all, rather than blocks encrypted under a schedule
  that was never expanded. NULL guards added to all five public functions.
- **`services/crypto/src/rsa_ecc_sha512.c`** — NULL guards on `eos_sha512_init`,
  `eos_sha512_update`, `eos_sha512_final` and `eos_sha512_hex`, mirroring `sha256.c` exactly.
- **`tests/test_crypto_aes.c`** — six new cases (10 → 16) covering each rejected key size, a
  negative size, a NULL key, and that 128/192/256 are still accepted.
- **`tests/CMakeLists.txt`** — register `test_crypto_aes` and `test_crypto_sha512`, which existed
  in the tree but were never built or run. `ctest` goes 21 → 23 suites.
- **`.github/workflows/ci.yml`** — add `master` to the `push` and `pull_request` triggers, and
  correct `-DBUILD_TESTS=OFF` to `-DEOS_BUILD_TESTS=OFF` in the `build-arm` and `release` jobs.

No signature changed and no header changed, so **this is not an API break**. The guard style
follows `9659598` (`fix: reject invalid dependency graph inputs`) — early returns at the public
entry point.

## Testing

Baseline recorded before any change: **21/21 ctest, 10/10 pytest, build exit 0.**

| Check | Command | Result |
|---|---|---|
| Build (Debug) | `cmake --build build/host -j8` | `PASS` — exit 0, **0 warnings** |
| Build (Release, clean tree) | `cmake -B build/clean -DEOS_BUILD_TESTS=ON && cmake --build` | `PASS` — exit 0 |
| Warnings, changed files | `gcc -Wall -Wextra -Werror -O2 -c` on all 4 changed `.c` | `PASS` — exit 0 on each |
| Unit tests | `ctest --output-on-failure --no-tests=error` | `PASS` — **23/23** (was 21/21) |
| Regression — break the code | new tests vs `git show HEAD:` sources | `PASS` — both **SIGSEGV, exit 139** |
| ASan + UBSan, key sizes | `-fsanitize=address,undefined -fno-sanitize-recover=all` | `PASS` — `HEAD` aborts on the first case; fixed tree returns cleanly on all 5 |
| ASan + UBSan, full suites | `test_crypto_aes`, `test_crypto_sha512` | `PASS` — 16/16 and 6/6, zero diagnostics |
| Valgrind | 4 crypto suites, `--leak-check=full` | `PASS` — `ERROR SUMMARY: 0 errors` on each |
| Python tests | `python3 -m pytest tests/ -q` | `PASS` — 10 passed (unchanged) |
| Product profiles | `robot`, `gateway`, `medical` | `PASS` — exit 0 each |
| Cross-compile ARM | — | `NOT RUN` — `gcc-arm-none-eabi` not installed in my environment |
| MSVC / macOS build | — | `NOT RUN` — no Windows or macOS compiler available to me |
| clang-tidy / cppcheck | — | `NOT RUN` — not executed locally |
| CI on this PR | — | `NOT RUN` — I have not seen a CI run for this branch |

Each new test asserts `nr == 0` **and** then drives all four public routines through a shared
helper that fails unless the output buffer is left byte-for-byte untouched — the fail-closed
property, not just the flag.

`test_aes_accepts_standard_key_sizes` guards the other direction: 128/192/256 still produce
`nr == 10/12/14`. AES-192 has no NIST vector anywhere in this suite (a pre-existing gap), so it is
covered structurally — `nr` plus an encrypt/decrypt round-trip — not against the standard. The
existing NIST FIPS-197 and SP 800-38A vectors in both files pass unchanged; the AES and SHA-512
arithmetic is untouched by this PR.

- [x] Unit tests pass (`ctest --test-dir build --output-on-failure`) — 23/23
- [ ] Integration tests pass — `NOT RUN`, this change is confined to one library
- [x] Manual testing performed — ASan/UBSan and Valgrind harnesses, output below
- [x] New tests added for new functionality — 6 cases, each confirmed failing against `HEAD`

## Pre-Submission Checklist

- [x] Code compiles without warnings (`-Wall -Wextra -Werror` for C) — verified at `-O2` on all
      four changed `.c` files; 122 pre-existing warnings elsewhere in the repo are listed under
      Additional Notes, none in files this PR touches
- [x] All existing tests pass — 23/23 ctest, 10/10 pytest
- [ ] Documentation updated if API changed — **no API change**, so nothing to update
- [x] Branch is rebased on latest `master` (`9659598`)

## Related Issues

None — found by inspection while reading the crypto service. Happy to open tracking issues for
any of the items under Additional Notes if that is preferred.

## Screenshots / Logs

**ASan against `HEAD`, `key_bits = 512`:**

```
services/crypto/src/aes.c:74:16: runtime error: index 60 out of bounds for type 'uint32_t [60]'
==393==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7468b3b00114
WRITE of size 4 at 0x7468b3b00114 thread T0
    #0 0x5b819532855d in eos_aes_init services/crypto/src/aes.c:74
Address 0x7468b3b00114 is located in stack of thread T0 at offset 276 in frame
  This frame has 2 object(s):
    [32, 276) 'ctx' <== Memory access at offset 276 overflows this variable
```

**The other two cases against `HEAD`:**

```
key_bits=0 :  services/crypto/src/aes.c:69:29: runtime error: index -1 out of bounds
              AddressSanitizer: stack-buffer-underflow ... READ of size 4

key_bits=64:  services/crypto/src/aes.c:71:56: runtime error: index 11 out of bounds
              for type 'uint8_t [11]'
              AddressSanitizer: global-buffer-overflow ... READ of size 1
```

**Break the code and watch it fail** (per `TESTING.md`). The new tests were compiled against the
pre-fix sources extracted with `git show HEAD:…`, without mutating the working tree:

```
===== test_crypto_aes (against HEAD sources) =====        EXIT=139   (SIGSEGV)
===== test_crypto_sha512 (against HEAD sources) =====     EXIT=139   (SIGSEGV)

===== test_crypto_aes (against working tree) =====   16/16 tests passed   EXIT=0
===== test_crypto_sha512 (against working tree) =====  6/6 tests passed   EXIT=0
```

Because the pre-existing `test_aes_null_safety` crashes first against `HEAD`, the six new
key-size cases were also exercised in isolation through a standalone harness, so their independent
failure is evidenced rather than assumed:

```
#  key_bits harness vs OLD (HEAD) aes.c
  key_bits=512    aes.c:74:16: runtime error: index 60 out of bounds    EXIT=1

#  key_bits harness vs NEW (fixed) aes.c
  key_bits=512    returned, nr=0
  key_bits=0      returned, nr=0
  key_bits=64     returned, nr=0
  key_bits=-128   returned, nr=0
  key=NULL        returned, nr=0
  ALL RETURNED CLEANLY                                                  EXIT=0
```

**Why `ci.yml` has never run** — it triggers on push to `main`/`develop` and PRs to `main`:

```
$ git ls-remote --heads upstream
f41201f…  refs/heads/chore/audit-and-ci
9659598…  refs/heads/master
ff48225…  refs/heads/release
```

No `main`. No `develop`. Every other workflow in `.github/workflows` targets `master`. This matters
because `ci.yml` is the **only** workflow that configures with `EOS_BUILD_TESTS=ON` and runs
`ctest` — `build.yml` hand-compiles four test files with `gcc` and never invokes `ctest`, so the
registered suite has had no CI coverage at all. `master` is **added** to both triggers rather than
replacing `main`, so nothing breaks if the default branch is later renamed.

### Scope and limitations — what I deliberately did NOT change

- **`eos_aes_init()` still returns `void`.** An `int` status would tell callers *why* a key was
  rejected, but that breaks the published API in `crypto.h`. The `nr == 0` sentinel gets the
  safety property without the break. A maintainer decision, not one to take inside a bug fix.
- **AES-CBC still silently ignores a trailing partial block.** `for (off = 0; off + 16 <= len; …)`
  leaves the final `len % 16` bytes of `out` untouched — uninitialised on encrypt. That is a
  padding-scheme design question (PKCS#7? reject? document as the caller's job?), not something to
  settle unilaterally. **Not fixed, and not covered by the tests added here.**
- **In-place CBC decryption is unsafe.** `eos_aes_cbc_decrypt` copies `prev` from `in + off`
  *after* writing `out + off`, so `in == out` corrupts the chain. No caller does this today and the
  header does not promise it — flagging rather than fixing.
- **`eos_aes_init` has no in-tree callers outside the tests and the fuzz harness.** This is a
  latent public-API memory-safety defect, not a reachable exploit in the current tree. Stating that
  plainly rather than overselling the severity.
- **14 other test files are also unregistered** — `test_e2e`, `test_stress`, `test_profiles`,
  `test_board_configs`, `test_driver_recovery`, `test_net_mock`, `test_net_integration`,
  `test_emulation_simulation`, `test_crypto_rsa`, `test_crypto_ecc` and 4 more. I built and ran
  them locally: **9 pass**, and 4 fail to build for environment reasons — `test_firmware`
  (undefined `eos_backend_find*`), `test_hal_stm32f4` (needs `-Iboards/stm32f407_discovery`),
  `test_performance_benchmarks` (`CLOCK_MONOTONIC` needs `_POSIX_C_SOURCE`), `main_system_test`
  (freestanding ARM target). Only the two tied to this fix are registered here, to keep the diff
  reviewable. The rest is a worthwhile follow-up I am happy to open separately.
- **122 pre-existing compiler warnings** in the Release build across 25 files (32 in
  `core/src/config.c` alone), mostly `-Wformat-truncation=`. **Zero are in files this PR touches.**
  Left alone, but reported per `CLAUDE.md`.
- **`.github/PULL_REQUEST_TEMPLATE.md` is corrupted.** It contains literal control bytes where
  letters belong — `0x0C` for `f` in "feat"/"fix", `0x0D` in "refactor", `0x09` in "test", `0x08`
  in "build" — so its Type of Change list renders as "eat", "ix", "efactor", "uild". Something
  interpreted backslash escapes when the file was written. I have written the labels correctly
  above. One-line fix, unrelated to crypto — say the word and I will send it separately.
- **`build_coverage/`, `build_sim/` and `build_qemu_arm64/` are committed** — 1,431 tracked build
  artifacts including `CMakeCache.txt` files carrying absolute paths. `.gitignore` covers `build/`
  and `build-*/` but not `build_*/`. Untouched: removing them is a large, noisy diff and a
  maintainer call.


srpatcha:
### Review — the key-size validation is still needed; the NULL guards have landed

Built merged onto current `master`: **conflict** in `services/crypto/src/aes.c` and `services/crypto/src/rsa_ecc_sha512.c`.

The reason is that #45 merged after you opened this and added the NULL guards independently. Current `master` has `if (!ctx || !key) return;` in `eos_aes_init` and `if (!ctx || !in || !out) return;` in both block functions.

**What is still missing on `master`, and is the valuable part of this PR:**

```c
void eos_aes_init(EosAesCtx *ctx, const uint8_t *key, int key_bits) {
    if (!ctx || !key) return;
    int nk = key_bits / 32;      /* key_bits is never validated */
    ctx->nr = nk + 6;
```

- `key_bits = 512` → `nk = 16`, `nr = 22`, `total = 92` words written into `ctx->rk[60]` — **a 32-word overflow past the end of the context**.
- `key_bits = 0` → `nk = 0` → `i % nk` is a **division by zero**.
- `key_bits = 64` → indexes `rcon[]` out of bounds.

Your `key_bits != 128 && != 192 && != 256` check fixes all three, and zeroing the context so `nr == 0` gives a fail-closed state that the block functions then refuse. That combination is the right design — reject at init, and make the rejected state unusable rather than merely uninitialised.

**To unblock:** rebase onto current `master` and keep only

1. the `key_bits` whitelist in `eos_aes_init`
2. the `ctx->nr == 0` guards in encrypt/decrypt/CBC
3. `tests/test_crypto_aes.c`

The NULL checks will drop out as already-applied. The `ci.yml` changes in this PR (`master` added to the branch triggers, `BUILD_TESTS` → `EOS_BUILD_TESTS`) are also correct and are being handled in #68 — safe to drop them here.

Worth knowing: your `-Werror`-clean rebase will now actually be compiled by CI once #68 lands. It has not been until now.

srpatcha:
## Review — the fix is right; here is the merge resolution it needs

Merged this onto current `master` locally. Two conflicts plus one that git did not flag.

### `services/crypto/src/aes.c` — take this PR's side

```c
<<<<<<< HEAD  (this PR)
    if (!key || (key_bits != 128 && key_bits != 192 && key_bits != 256)) {
        memset(ctx, 0, sizeof(*ctx));
        return;
    }
=======       (master)
    if (!ctx || !key) return;
>>>>>>>
```

Master accepts any `key_bits` and leaves `ctx` holding whatever the caller passed in. This PR rejects the invalid sizes and zeroes the context so a rejected init cannot be mistaken for a usable one. The `ctx->nr == 0` guards added to encrypt/decrypt are the other half of that — they are what makes the zeroing observable instead of silently encrypting with an uninitialised key schedule. Strictly stronger; keep it.

### `services/crypto/src/rsa_ecc_sha512.c` — take master's side

Three of the four hunks are a blank line. The fourth is real but equivalent:

```c
this PR   if (!ctx || !data || len == 0) return;
master    if (!ctx || (!data && len > 0)) return;
```

Both no-op on `len == 0` and both reject `data == NULL`, so behaviour matches. Master's reads better for a hash update, where `update(ctx, NULL, 0)` is a legitimate call rather than an error, so I would keep master's here and drop the whitespace churn.

### `tests/CMakeLists.txt` — git merges this cleanly and the result does not configure

Not reported as a conflict, so it is easy to miss:

```
CMake Error at tests/CMakeLists.txt:67 (add_test):
  add_test given test NAME "test_crypto_sha512" which already exists in this directory.
```

Both sides added registrations for `test_crypto_aes` and `test_crypto_sha512`; the textual merge kept both copies. Deleting the duplicated block at lines 59–67 is the whole fix.

### After those three

```
cmake configure  OK
build            no errors
ctest            100% tests passed, 0 tests failed out of 24
```

Nothing here needs a code change from you beyond the rebase — the crypto hardening is sound and I would like it in. Flagging the `tests/CMakeLists.txt` duplicate specifically because CI would have hit it after a green-looking merge.

### Verification

Local merge onto `origin/master`, conflicts resolved as above, then a full cmake + ninja + ctest run. Not run: anything on real hardware.

srpatcha (dismissed):

srpatcha (approved):

PR #60, opened 2026-08-26
fix: improve Windows host build and strengthen heap/package tests

Rani2909:
## Summary

This PR improves Windows host/simulation compatibility and strengthens the heap and package test coverage.

### Changes

* Removed the x64 `hlt` instruction from the host/simulation idle-task path in `kernel/src/task.c` to avoid executing a privileged CPU instruction during host-based testing.
* Expanded `tests/test_heap.c` to validate:

  * Basic allocation
  * 8-byte alignment
  * Memory read/write
  * Multiple allocations
  * Free and re-allocation
  * `calloc` zero-initialization
  * `realloc` data preservation
  * Heap statistics
  * Double-free protection
* Updated `tests/test_package.c` to use static `EosPackageSet` instances, preventing stack-related issues in the Windows test environment.

### Validation

* Windows/MSVC Debug build completed successfully.
* All 21 CTest tests passed.
* Test result: **21/21 passed (100%)**
* `git diff --check` completed with no errors.

### Test Command

```text
ctest --test-dir build_sim_windows -C Debug --output-on-failure
```

### Result

```text
100% tests passed out of 21
```

### DIFF
--- a/tests/test_heap.c
+++ b/tests/test_heap.c
@@ -175,6 +175,7 @@ int main(void)
     eos_free(a);
     eos_free(c);
     eos_free(d);
+    eos_free(zero);
     eos_free(r2);
 
     eos_heap_stats(&stats);
--- a/tests/test_heap.c
+++ b/tests/test_heap.c
@@ -14,11 +14,19 @@ static uint8_t foreign[128];
 
 int main(void)
 {
+    eos_heap_stats_t stats;
+
+    /* ---------------------------------------------------------
+     * 1. Heap initialization
+     * --------------------------------------------------------- */
     if (eos_heap_init(heap_area, sizeof(heap_area)) != 0) {
         fprintf(stderr, "heap initialization failed\n");
         return 1;
     }
 
+    /* ---------------------------------------------------------
+     * 2. Overflow protection
+     * --------------------------------------------------------- */
     if (eos_malloc(SIZE_MAX) != NULL) {
         fprintf(stderr, "SIZE_MAX allocation must be rejected\n");
         return 1;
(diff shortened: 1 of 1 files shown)


srpatcha:
### Review — three separate changes; two are fine, one needs a decision

**1. `test_package.c` static sets** — correct, and byte-identical to #52 and to the same hunk in #48 (compared blob hashes). #48 carries the explanatory comment, so that is probably the one to land; this hunk then rebases to nothing. No action needed from you.

**2. `test_heap.c` expansion** — mostly a clear improvement: alignment, write/read-back, calloc zeroing, realloc preserving data, and a stats check. One part I'd ask you to reconsider:

```c
/* 11. Double-free should not corrupt allocator */
eos_free(a);          /* a was already freed in step 9 */
```

This pins double-free tolerance as a **contract**. If the allocator is ever changed to detect double-free and trap — which is a defensible hardening choice — this test starts failing and someone will "fix" it by weakening the allocator. If the intent is "we currently survive this", a comment saying so keeps the option open. If the intent is "we guarantee this", it should be stated in the heap's documentation, because it constrains future work.

**3. Removing the x86_64 `hlt` from `idle_task_func`** — this is the one that needs a decision, and it is not really a test change:

```c
-#elif defined(__x86_64__) || defined(_M_X64)
-    __asm volatile ("hlt");
```

You are right that `hlt` is privileged and faults in a user-space host build. But this also removes it for a genuine bare-metal x86_64 target, where it is the correct instruction and the power saving is the point. **Gate it rather than delete it** — on `EOS_PLATFORM`, on a hosted/freestanding check, or on `__STDC_HOSTED__` — so the host build stops faulting without silently making bare-metal x86 spin at 100%.

Splitting 3 into its own PR would let 1 and 2 land quickly.

Rani2909:
Thanks for the review. I’ve addressed the requested changes:

* Removed the double-free tolerance test from `test_heap.c` so the test does not establish a future allocator contract.
* Kept the `test_package.c` static `EosPackageSet` changes as requested.
* Split the x86_64 `hlt` change into separate PR #76 so the hosted-build fix can be reviewed independently.

PR #60 now contains only the heap and package test improvements. The working tree is clean and the branch is pushed to my fork.

srpatcha:
## Needs a rebase — #77 landed in the same file

Only `tests/test_heap.c` conflicts. #77 merged a heap-header validation change that added realloc coverage to the same file, so the two sets of additions collide textually.

Worth reading #77 before you resolve, because it overlaps this PR's subject rather than just its line numbers. It added:

- `block_from_ptr()`, which range-checks and sanity-checks a header, now used by **both** `eos_free` and `eos_realloc` — previously only `eos_free` validated, so `eos_realloc` trusted whatever it was handed.
- A rejection for headers claiming a size below `HEADER_SIZE`, because `block->size - HEADER_SIZE` is unsigned and underflowed to near `SIZE_MAX`, making every growth request look already satisfied.
- Tests for a foreign pointer, an undersized header, and payload preservation across a growing realloc.

So some of what this PR strengthens in `test_heap.c` may now be covered. Keep whatever is genuinely additional and drop the rest — the Windows host-build half of this PR is untouched by #77 and is the part I am most interested in, since none of it can be exercised on this machine.

On that: I cannot verify the Windows changes here at all. Linux is the only host available to me, so for that half I am relying on your testing rather than mine, and I would rather say so than imply I checked it. If CI covers the Windows build after the rebase, that is the evidence I would go on.

`tests/test_package.c` also changed under you via #48 (all `EosPackageSet` locals became `static`, with a comment explaining the ~3.8 MB against a 1 MB Windows stack) — that one auto-merges, but it is adjacent to what you are doing and may already solve part of it.

Rani2909:
Thanks for the review. I’ve completed the requested rebase and cleanup.

* Rebased PR #60 onto the current `master`, including the changes from #77.
* Removed the overlapping realloc coverage where it was already provided by #77.
* Removed the double-free test so it does not establish a future allocator contract.
* The `test_package.c` change is now already covered by the merged upstream fix.
* The x86_64 `hlt` change has been separated into PR #76 as requested.
* PR #60 now contains only the additional heap test coverage.

The branch is clean and pushed to my fork. I’ve also verified that the current PR diff contains only `tests/test_heap.c`.

Please take another look when convenient.

srpatcha (approved):

PR #61, opened 2026-08-26
fix(parser): out-of-bounds write in config parser when package count exceeds EOS_MAX_PACKAGES

alamrezoanul:
### Issue

`eos_config_load()` in `core/src/config.c` parses a YAML-like config file with
a hand-rolled line-by-line state machine. Every time it sees a new `- name:
...` entry under `packages:`, it does:

```c
pkg_idx++;
if (pkg_idx < EOS_MAX_PACKAGES) {
    strncpy(cfg->packages[pkg_idx].name, val, EOS_MAX_NAME - 1);
    cfg->package_count++;
    section = SEC_PKG_ENTRY;
}
```

If `pkg_idx` reaches `EOS_MAX_PACKAGES` (128), the `strncpy`/`package_count++`
are correctly skipped — but `section` is **not** reset. It stays at whatever
it was, and on the next few lines the parser falls into `case SEC_PKG_ENTRY:`,
`case SEC_PKG_BUILD:`, or `case SEC_PKG_OPTIONS:`, each of which writes
directly through `cfg->packages[pkg_idx]` with no bounds check:

```c
case SEC_PKG_ENTRY:
    if (strcmp(key, "version") == 0) strncpy(cfg->packages[pkg_idx].version, val, ...);
    ...
```

With `pkg_idx == EOS_MAX_PACKAGES`, `cfg->packages[pkg_idx]` reads one element
past the end of the `packages[]` array. Looking at `EosConfig` in
`core/include/eos/config.h`, the very next field in the struct after
`packages[EOS_MAX_PACKAGES]` is `int package_count`. So a config file with
more than 128 packages silently corrupts `package_count` (and adjacent
memory) with attacker/author-controlled bytes from the 129th+ package's
`version`, `source`, `hash`, build type, or option strings.

`eos_config_load()` is the first thing called on any `eos.yaml` project file
— this is reachable just by pointing the tool at a config with a long
`packages:` list, no special build options required.

### Approach

Two changes in `core/src/config.c`, both inside `eos_config_load()`:

1. When `pkg_idx >= EOS_MAX_PACKAGES` is detected at the point a new package
   name is parsed, log a clear error (`"Too many packages in config (max
   128); ignoring '<name>'"`) and explicitly reset `section = SEC_PACKAGES`.
   This means the following `key: value` lines for the rejected package are
   parsed as loose top-level lines and simply ignored, instead of being
   routed into the out-of-bounds `SEC_PKG_ENTRY`/`SEC_PKG_BUILD`/
   `SEC_PKG_OPTIONS` cases.
2. As defense in depth, added an explicit `pkg_idx < 0 || pkg_idx >=
   EOS_MAX_PACKAGES` bounds check at the top of `SEC_PKG_ENTRY` and
   `SEC_PKG_BUILD`, and extended the existing (already-partial) bounds check
   in `SEC_PKG_OPTIONS` to also check the upper bound, not just `pkg_idx >=
   0`. This means even if some other code path in the future changes
   `section` without going through the name-parsing branch above, these
   three cases can never index `packages[]` out of range.

Behavior change: a config with more than `EOS_MAX_PACKAGES` packages now
loads successfully, keeps the first 128 packages intact, and logs an error
for each package past the limit instead of corrupting memory. This seemed
better than hard-failing the whole load, since it matches how the rest of
the parser already treats other per-entry limits (e.g. `EOS_MAX_DEPS`,
`EOS_MAX_OPTIONS` are silently capped, not fatal).

### Testing

Added `test_config_load_package_overflow()` to `tests/test_config.c`:
generates a temp YAML config with `EOS_MAX_PACKAGES + 3` (131) packages, each
with a name, version, build type, and one option, then asserts:

- `eos_config_load()` still returns `EOS_OK`.
- `cfg->package_count == EOS_MAX_PACKAGES` (capped, not corrupted, not 131).
- The first package's fields are intact.
- The **last in-bounds** package (index 127) has its own name/version/
  option_count — i.e. it wasn't overwritten by data bleeding in from the
  rejected 129th–131st packages.

Verified the test is meaningful, not vacuous:
- Reverted only the `config.c` fix (kept the new test) and reran — the
  unpatched binary **segfaults** on this test (confirmed via `git stash`
  before/after).
- With the fix applied: new test passes 28/28 internal assertions.
- Full existing suite: `ctest` reports 21/21 tests passing — no regressions.

### Limitations / follow-ups

- This fix caps and logs; it does not turn "too many packages" into a hard
  config-load failure. If the project prefers strict validation (reject the
  whole file) that's a small follow-up but is a behavior/API decision I
  didn't want to make unilaterally in a bug-fix PR.
- I only audited the `packages:` list-parsing path (the one with the
  reported OOB write). I did not do a full audit of every other section of
  `eos_config_load()` for similar unbounded-index patterns — worth a
  follow-up pass if the maintainers want it.


srpatcha:

Built this merged onto current `master` and ran the suite: **24/24 ctest pass, no new warnings.**

The important evidence: I ran **your regression test against unfixed `master`** and it **segfaults** (exit 139). That confirms the out-of-bounds write is real memory corruption, not a theoretical bound. Thank you for writing a test that actually reproduces it — that is what made this reviewable in minutes rather than hours.

The fix itself is right. Falling back to `SEC_PACKAGES` rather than leaving `pkg_idx` dangling is the correct choice, because the `key: val` lines that follow a rejected package would otherwise be dispatched into `SEC_PKG_ENTRY`/`SEC_PKG_BUILD`/`SEC_PKG_OPTIONS` and index past the array. The defensive bounds in the `switch` are worth keeping even though the fallback makes them unreachable today.

**One note for the merge order:** this conflicts with #46 in `tests/test_config.c` — both of you append a test function at the same place. Trivial to resolve. I'd suggest merging **this one first** since it fixes memory corruption, and #46 rebases on top.

Also relevant: #65 fixes the same defect independently. Please don't take that as a reason to hold this — see my comment there.

alamrezoanul:
Hi @srpatcha , 

Do I have to do anything? Or someone else will do it?

Regards,
Rez.

alamrezoanul:
@srpatcha - asking because I couldn't understand whom you have told it....

srpatcha:
## Correct diagnosis — needs a rebase, and there is an overlap to settle

You and #65 independently found the same bug, and I confirmed it is real and reachable on `master` with AddressSanitizer using an ordinary 129-package config — no malformed input needed:

```
==ERROR: AddressSanitizer: stack-buffer-overflow
WRITE of size 127 at 0x7ffd1c4242b0 thread T0
    #1 in eos_config_load core/src/config.c:247
```

### The rebase

Only `tests/test_config.c` conflicts, from #46 landing `test_lockfile_freshness()` in the same region. Keep both test functions and drop the markers; `main()` merges cleanly on its own.

### The overlap

I have **#78** open against the same code. It came out of reviewing #65, which turned out not to compile, so I wrote the fix separately and extended it to four write sites the original patch did not reach — the keyed writes under `SEC_PKG_ENTRY`, `SEC_PKG_BUILD` and `SEC_PKG_OPTIONS`, which are the ones the ASan trace actually lands on (`config.c:247`).

Worth comparing yours against that before you rebase: if #61 covers only the `- name:` entry point, the `version:`/`source:`/`hash:` lines under a refused package still write through the stale index and the overflow remains reachable. If yours does cover them, I would rather merge yours and close #78 — you filed first.

One thing from writing the regression test that will save you time either way: **asserting on `package_count` and the accepted entries does not catch this**. Both are correct while the overflow is happening, and my first attempt at a test passed against the unfixed parser. #78 puts the config in a struct with a trailing canary so the stray write is observable — it reports 1150 clobbered bytes and exits 1 against unfixed code. Worth borrowing that shape rather than rediscovering it.

Tell me which way you want to go and I will make it happen.

alamrezoanul:
@srpatcha - so far I understood, we should rebase

alamrezoanul:
Hi @srpatcha - However I've resolved the conflict now as you have said to accept both changes. Could you please check and confirm ?

srpatcha (dismissed):

srpatcha (approved):

PR #64, opened 2026-08-26
fix(kernel): return invalid task name for deleted task handles

bhavyandaiya:
ISSUE:
In `kernel/src/task.c`, `eos_task_get_name()` previously only verified if the handle index was within bounds and if `g_tasks[h].name` was non-null. When a task was deleted via `eos_task_delete()`, its name pointer remained intact in the TCB while its state transitioned to `EOS_TASK_DELETED`. Calling `eos_task_get_name()` on a deleted task handle incorrectly returned the stale name instead of `"invalid"`.

APPROACH:
- Updated `eos_task_get_name()` in `kernel/src/task.c` to check `g_tasks[h].state == EOS_TASK_DELETED` alongside bounds and null checks, safely returning `"invalid"` for terminated tasks.
- Updated `test_task_delete()` in `tests/test_kernel.c` with assertions confirming `eos_task_get_name()` resolves to the task name before deletion and transitions to `"invalid"` following `eos_task_delete()`.

TESTING & VALIDATION PERFORMED:
- Built host test binaries using CMake (`cmake -B build/host -DEOS_BUILD_TESTS=ON && cmake --build build/host --parallel`).
- Executed full test suite via CTest: 21/21 tests passed (100% pass rate).
- Verified `test_kernel` passes with the new post-deletion assertion.

LIMITATIONS:
- Non-breaking change ensuring robust state validation for deleted tasks.

### DIFF
--- a/kernel/src/task.c
+++ b/kernel/src/task.c
@@ -360,7 +360,9 @@ eos_task_state_t eos_task_get_state(eos_task_handle_t h)
 
 const char *eos_task_get_name(eos_task_handle_t h)
 {
-    if (h >= EOS_MAX_TASKS || !g_tasks[h].name) return "invalid";
+    if (h >= EOS_MAX_TASKS || g_tasks[h].state == EOS_TASK_DELETED || !g_tasks[h].name) {
+        return "invalid";
+    }
     return g_tasks[h].name;
 }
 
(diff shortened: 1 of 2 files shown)


srpatcha:

Small, correct, and the test asserts both directions — the name before deletion and `"invalid"` after. That is what makes it a real regression test rather than a snapshot.

srpatcha (approved):

Correct: `eos_task_delete` leaves `name` populated, so the handle kept resolving to the deleted task's name and a caller had no way to tell a live task from a freed slot. Checking `EOS_TASK_DELETED` alongside the existing bound is consistent with what `eos_task_get_state` already reports. The test asserting the name both before and after the delete is what makes it a real guard rather than a restatement of the new branch.

PR #66, opened 2026-08-27
feat: add HiLetGo ESP32 board support

vivek3210:
## Issue, Bug, or Improvement

Add board support metadata for the HiLetGo ESP-WROOM-32 ESP32 ESP-32S module so it can be selected and validated as an ESP32 Xtensa LX6 board configuration.

## Proposed Approach and Implementation

- Added boards/hiletgo-esp-wroom-32.yaml
- Avoided advertising native USB; these boards typically use an external USB-to-UART bridge
- Added strict descriptor-specific validation for required YAML records, indentation, duplicates, schema values, peripheral mappings, and native USB absence
- Registered the board validation executable with CTest
- Documented that this repository does not currently contain an Xtensa firmware build target or flashing integration

## Testing and Validation

- PASS: Independent code review.
- PASS: Descriptor/schema contract checks by inspection.
- PASS: git diff --check.

## Additional Considerations and Limitations

- The descriptor targets the common ESP32-D0WDQ6 ESP-WROOM-32 profile; module flash size and exact board revision should be confirmed against the physical board before flashing.
- This PR adds board metadata and validation, not a complete ESP-IDF-compatible firmware port or flashing workflow.
- The PR is submitted from a fork because the authenticated account does not have write permission to the upstream repository.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -158,6 +158,14 @@ nrfjprog --reset
 esptool.py --chip esp32 write_flash 0x10000 build/eos.bin
 ```
 
+The HiLetGo ESP-WROOM-32 ESP-32S is represented by the
+`boards/hiletgo-esp-wroom-32.yaml` descriptor and reuses the existing ESP32
+(`ESP32-D0WDQ6`, LX6) target definition. EoS does not currently provide an
+Xtensa ESP32 cross-toolchain file or an ESP32 firmware build target, so this
+descriptor does not by itself produce a flashable image. Install and configure
+the Espressif ESP32 toolchain and validate the generated image and flash
+addresses for the specific module before programming hardware.
+
 ---
 
 ## Build eBoot Separately
--- a/boards/hiletgo-esp-wroom-32.yaml
+++ b/boards/hiletgo-esp-wroom-32.yaml
@@ -1,0 +1,33 @@
+# EoS Board: HiLetGo ESP-WROOM-32 ESP32 ESP-32S
+board:
+  name: hiletgo-esp-wroom-32
+  mcu: ESP32-D0WDQ6
+  family: ESP32
+  arch: xtensa
+  core: lx6
+  vendor: Espressif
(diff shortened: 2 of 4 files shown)


srpatcha:

Built merged onto current `master`: **25/25 ctest pass** (your board-config test brings it from 24).

Two things I appreciated:

1. **The honest DEPLOYMENT.md note.** Stating plainly that EoS has no Xtensa toolchain or ESP32 firmware target, so the descriptor does not by itself produce a flashable image, is exactly right. §25 of the architecture document exists to prevent the opposite.
2. **You registered `test_board_configs`.** That file existed in `tests/` and no `CMakeLists.txt` referenced it, so it had never been compiled or run. It is one of 13 such files on `master`.

Two small things:

- `boards/hiletgo-esp-wroom-32.yaml` has **no trailing newline** at EOF.
- The board is described as reusing the `ESP32-D0WDQ6` definition. Worth a line in the YAML recording which existing target it derives from, so the relationship survives someone editing one of them.

Relevant context: the adoption plan proposes picking ~3 validated reference boards rather than adding descriptors. This one is different from the 85 `generic-*` entries because it has a test behind it and an honest status note — that is the standard the others should be held to.

srpatcha (approved):
Verified locally against current master: merges clean, builds with no new errors, `ctest` 26/26 (this adds one).

The board YAML is additive, so the interesting part is `test_board_configs`, and I like that it validates *every* file in `boards/` rather than only the one being added — a board definition is exactly the kind of file that gets copied and half-edited, and this turns that into a test failure instead of a boot failure. Skipping comment lines in `file_contains_text` matters more than it looks: without it a commented-out key satisfies its own check.

One thing to be aware of rather than change: `WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}` ties the test to the source layout, so it will not run from an installed or packaged tree. That is the right trade for a config-validation test — it has to read the real files — but if boards ever ship as data, this is the line that needs revisiting.

PR #67, opened 2026-08-27
Correct and transitive mutex priority inheritance

shre111:
## What this changes

Rewrites how the mutex implementation handles priority inheritance. There is also a small unrelated test fix I needed in order to get the suite running at all.

## The problem

`kernel/src/sync.c` did priority inheritance by saving the owner's priority into the mutex when it was locked, then writing that value back when it was unlocked. That works for the simple case of one mutex, one waiter and no timeouts. Outside that case it goes wrong in ways that either reintroduce priority inversion or corrupt a task's priority permanently.

I found six separate problems.

**1. Releasing one of several held mutexes dropped every boost.** `original_prio` was stored per mutex. A task holding M1 and M2, boosted because a high priority task is waiting on M2, would fall back to its pre M1 priority as soon as it released M1. The high priority task is still blocked on M2 at that point, so the inversion is immediately back.

**2. A boost taken while already boosted became the task's new baseline.** If a task locks a second mutex while it is inheriting, the inherited priority gets recorded as that mutex's "original". Release everything and the task stays at the elevated priority for good, starving every task in between.

**3. A waiter that timed out left its boost behind.** The timeout path removed the waiter from the queue but never recomputed the owner, so the owner kept the inherited priority even though nothing was waiting on it any more.

**4. Boosts stopped at the immediate owner.** If A blocks on a mutex held by B, and B is itself blocked on a mutex held by C, only B got boosted. C is the task that actually has to run before anything can be released, and it stayed at its low priority, so A effectively waited at C's priority. This is the classic chained inversion, the same shape as the bug that kept resetting Mars Pathfinder.

**5. `eos_mutex_delete` leaked the boost.** It woke the waiters but never recomputed the owner.

**6. A full wait queue blocked a task that was never queued.** The code only added the caller to the waiter array if there was room, but blocked it either way. `eos_mutex_unlock` only ever hands the mutex to a queued waiter, so with `EOS_WAIT_FOREVER` such a task sleeps forever. `eos_sem_wait` had the same problem.

## What I did

Replaced the save and restore approach with an invariant that gets recomputed whenever anything changes:

```
effective = min( base_priority,
                 priority of every task waiting on every mutex this task holds )
```

with lower numbers meaning more urgent.

`base_priority` is a new field on the TCB, set once at task creation. That means inheritance always restores down to the priority the task was actually created with, and can never mistake a temporary boost for a baseline, which covers problem 2. Recomputing from the live set of held mutexes means releasing one keeps whatever the others still justify, covering problem 1, and running the same recomputation on the timeout, unlock and delete paths covers problems 3 and 5.

Since a waiter's own priority may itself be inherited, the recomputation walks along the chain of blocked tasks and pushes the boost through to whoever sits at the head of it. That covers problem 4.

The walk is bounded by `PI_MAX_CHAIN_DEPTH`, which is 8. The bound is not just tidiness. If application code takes two mutexes in opposite orders the chain becomes a cycle, and an unbounded walk would spin inside a critical section with interrupts masked. Worst case is `EOS_MAX_MUTEXES * MTX_MAX_WAITERS * PI_MAX_CHAIN_DEPTH`, which keeps the operation analysable for WCET.

Wait queue overflow now returns `EOS_KERN_NO_MEMORY` from both `eos_mutex_lock` and `eos_sem_wait`, instead of blocking a task that is not in the queue. That covers problem 6.

## API notes

* `eos_task_t` gains `base_priority`. The struct is public but the field is appended at the end, so this stays source compatible. `priority` keeps its existing meaning as the effective scheduling priority.
* `eos_task_set_current_internal()` is added to the internal header, not the public one. The Cortex-M port establishes the running task from PendSV via `eos_schedule()`. Ports without a hardware context switcher, and host builds, need an explicit way to say which task is running, otherwise anything that depends on the caller, mutex ownership most of all, cannot work or be tested.

No public API changed.

## Testing

Added `tests/test_priority_inherit.c` with six cases, one per problem above, wired into ctest.

There is a wrinkle worth explaining. On the host, `eos_port_yield()` is a stub that does nothing, so a contended lock falls straight through to its timeout path and removes the waiter again. The boosted state therefore only exists during the yield. Since the test file owns the yield stub, that is where the assertions go: they run while the waiter is still parked and the owner is still boosted. Nesting that hook is also how the queue overflow case parks enough waiters to fill the queue.

To check that the tests actually pin the behaviour rather than just passing, I compiled each one against both the previous `sync.c` (pulled out with `git show`) and the new one:

```
TEST                                          OLD    NEW
boost_survives_partial_release                FAIL   PASS
base_priority_not_overwritten_by_boost        FAIL   PASS
timeout_withdraws_boost                       FAIL   PASS
transitive_chain                              FAIL   PASS
delete_withdraws_boost                        FAIL   PASS
waitqueue_overflow_reports_error              FAIL   PASS
```

Full suite: ctest 22/22 pass, which is the 21 existing tests plus the new one.

The unrelated fix: `tests/test_package.c` declared `EosPackageSet` as a plain local in six functions. That struct is roughly 3.9 MB, so it overflows the default stack the moment the function is entered, and the test crashed with no output at all. `cmd/eos/main.c` already stores the same struct as `static` for exactly this reason. I changed the six declarations to match, and the test went from crashing to passing.

## Limitations

* Priority inheritance bounds inversion, it does not prevent deadlock. Taking two mutexes in opposite orders in two tasks still deadlocks. The depth bound only stops the kernel spinning while it happens. I noted this in the docs.
* I did not implement priority ceiling emulation. Inheritance alone was the scope here.
* The host tests exercise the priority bookkeeping, not real preemption. Checking actual scheduling behaviour under a real context switcher needs hardware or QEMU, and this change does not add that.

## Docs

Added a priority inheritance section to `docs/book/part2-kernel-hal/ch06-kernel.md` covering the rule, the transitive behaviour, the chain bound and the deadlock caveat. Noted the new error code and linked to that section from `docs/api-reference.md`.


srpatcha:
### Review — verified, ready to merge; the strongest PR in the queue

Built merged onto current `master`: **25/25 ctest pass** (your `test_priority_inherit` brings it from 24).

This is the right design, and specifically for the reason your docs give: recomputing

```
effective = min(base_priority, priority of every waiter on every mutex held)
```

on every change, instead of saving and restoring a value around each lock, is what makes the nested cases correct. A task holding two mutexes that releases one keeps the boost the other still justifies — the save/restore approach cannot express that.

Things I checked specifically:

- **`base_priority` cannot desync.** There is no public `eos_task_set_priority`, only the internal one, so nothing outside the kernel can change a task's priority behind the invariant. Worth a comment saying so, because that safety disappears the moment such an API is added.
- **`PI_MAX_CHAIN_DEPTH` terminates on a lock-order cycle.** The bound is doing real work, not decoration, and your docs are honest that it bounds inversion without preventing deadlock.
- **`pi_propagate` after ownership moves in `unlock`**, not before, so the releasing task keeps what its other mutexes justify. Easy to get backwards.
- **`eos_mutex_delete` retires the mutex before recomputing** so it no longer contributes. Also easy to get wrong.

The `EOS_KERN_NO_MEMORY` return when the wait queue is full is a real bug fix hiding inside a feature PR — a caller that was never enqueued would sleep forever under `EOS_WAIT_FOREVER`. Same change in `eos_sem_wait`. Both correct.

Two notes:

1. **`eos_task_set_priority_internal` is now unused by `sync.c`** (`pi_recompute` writes `g_tasks[].priority` directly). It is still exported in `kernel_internal.h`. Not a defect, but worth a look — two ways to set the same field is how invariants get bypassed later.
2. **The `test_package.c` hunk** duplicates #48/#52/#60. Harmless; it rebases to nothing once one of them lands.

#55 fixes the timeout case this also covers, and the two conflict in `sync.c`. See my comment there — that PR does not need closing, it needs a rebase decision.

srpatcha (dismissed):
Verified locally against current master: merges clean, builds with no new errors, `ctest` 25/25 (this adds one).

The design decision that makes this correct is recomputing

    effective = min(base_priority, priority of every waiter on every mutex held)

on each lock, unlock, timeout and delete, rather than saving a priority on the way in and restoring it on the way out. Save/restore is the usual implementation and it is wrong in exactly the cases that matter: a task holding two mutexes that releases one drops the boost the other still justifies, and a task boosted while already holding a mutex restores the inherited value as if it were its baseline. Keeping `base_priority` immutable in the TCB and deriving everything else from it removes both.

Transitivity is the other half and is genuinely needed — if A waits on B and B is blocked on C, C is the task that has to run before anything releases, so boosting only B accomplishes nothing. Bounding the walk with `PI_MAX_CHAIN_DEPTH` is right for two reasons and the comment names both: WCET analysability, and termination when an application creates a lock-order cycle.

Also correct to return `EOS_KERN_NO_MEMORY` rather than blocking when the wait queue is full. A task that blocks without being enqueued is never granted the mutex, so failing the call is the only honest outcome.

The documentation is unusually good — stating plainly that this bounds inversion but does not prevent deadlock, and that nested acquisition still needs a global lock order, is the thing most implementations leave for someone to discover.

srpatcha:
## Approved — one trivial rebase needed

Review is above; this is ready on the merits. #48 landed a few minutes ago and touched `tests/test_package.c`, which this branch also touches, so there is now exactly one conflict — and it is two comments saying the same thing:

```
<<<<<<< HEAD                                    ← yours
/* sizeof(EosPackageSet) is ~3.9 MB, which overflows the default stack as a
 * plain local. cmd/eos/main.c stores it static for the same reason. */
=======                                         ← master, from #48
/* EosPackageSet is EOS_MAX_PACKAGES * sizeof(EosPackage) = ~3.8 MB, which
 * overflows the 1 MB default thread stack on Windows ... */
>>>>>>> origin/master
```

You and #48 independently made the same `static` change for the same reason. Keep master's version of the comment — it is the longer one and names the Windows 1 MB stack specifically — and drop yours. The code lines are identical on both sides, so that single hunk is the whole rebase; `kernel/src/task.c`, `kernel/include/eos/kernel_internal.h` and `tests/CMakeLists.txt` all auto-merge clean.

Worth knowing what else moved under you while this was open, since both touch the kernel:

- **#63** (tick wraparound) landed and changed `kernel_internal.h` and `task.c`. It adds `EOS_TICK_REACHED` and a `wake_armed` flag on `eos_task_t`. That merged cleanly with your branch locally, but it is adjacent to your timeout path — worth a look at `eos_mutex_lock`'s timeout handling after the rebase to confirm the two agree on what an expired deadline means.
- **#64** made `eos_task_get_name` report `DELETED` handles as invalid.

I verified the rebase locally (resolving that one hunk as described) and it builds with `ctest` 25/25. Push the rebase and I will merge it.

PR #74, opened 2026-08-28
fix: handle package transitions after deps and build

muhammadburhandevv-hub:
## Summary

Fixes the eOS configuration parser so packages and dependencies are parsed correctly after `deps` and `build` sections.

Previously, a package following another package's dependency list could be treated as another dependency. A package-level `deps:` section following `build:` could also be ignored.

## Type of Change

- [ ] feat — New feature
- [x] fix — Bug fix
- [x] test — Add or fix tests

## Changes

- Recognize `- name:` as a new package while the parser is processing another package's dependency list.
- Continue treating normal list items inside `deps` as dependencies of the current package.
- Allow a package-level `deps:` section to follow its `build:` section.
- Add regression tests for a package following a dependency list and dependencies following a build section.

## Testing

- [ ] Unit tests pass (`ctest --test-dir build --output-on-failure`)
- [ ] Integration tests pass

Focused validation performed with MSVC on Windows:

```powershell
cmake -S . -B build\contributing-check -G "Visual Studio 17 2022" -A x64 -DEOS_BUILD_TESTS=ON -DEOS_PRODUCT=vbox_test
cmake --build build\contributing-check --config Release --target test_config --parallel
ctest --test-dir build\contributing-check -C Release -R "^test_config$" --output-on-failure --no-tests=error
```

Results:

- The configuration parser test passed: 1/1.
- All parser assertions passed: 34/34.
- Both regression tests were confirmed to fail when their corresponding parser changes were temporarily removed.

The full unit-test checkbox remains unchecked because the complete Windows build is blocked by an unrelated existing compiler error before every test executable can be built.

## Pre-Submission Checklist

- [ ] Code compiles without warnings (`-Wall -Wextra -Werror` for C)
- [ ] All existing tests pass
- [x] Documentation updated if API changed — not applicable because no public API changed

The changed parser and test target compile successfully. The complete repository build still contains unrelated existing warnings and an MSVC error. The branch was created from the local `master`, but it has not been verified against the latest upstream `master`.

## Related Issues

No issue number was provided.

## Screenshots / Logs

Focused CTest result:

```text
1/1 Test #1: test_config ...................... Passed
100% tests passed out of 1
```

Direct parser test result:

```text
=== Results: 34/34 passed ===
```

## Additional Notes

- No external dependencies were added.
- The existing YAML configuration format was not changed.
- The full Windows Release build is currently blocked by GCC-style `__asm("wfi")` in `kernel/src/task.c:175`, which MSVC x64 cannot compile.
- The full build also reports existing warnings in files unrelated to this PR.
- Linux, macOS, GCC, Clang, and embedded cross-compilation were not tested locally and should be validated by CI.


srpatcha:
## The fix is correct — it needs a rebase, and the conflict is trivial

Reviewed the parser change and it is right. A `- name:` item appearing while `section == SEC_PKG_DEPS` was falling into the dependency branch, so the package that followed a `deps:` list was swallowed as a dependency of the previous one:

```yaml
packages:
  - name: application
    deps:
      - logging-library
  - name: network-library   # ← became a dep of `application`
```

Adding `SEC_PKG_DEPS` to the outer condition and making the dependency branch the `else` of the `name` check gets the precedence the right way round: a `name` key always starts a package, anything else in that section is a dependency. The `deps:` transition added under `SEC_PKG_BUILD` closes the matching hole for `build:` followed by `deps:`.

### The conflict

`tests/test_config.c` now conflicts with master — #46 landed `test_lockfile_freshness()` in the same spot. It is a pure both-added-a-function conflict, not a semantic one:

```
<<<<<<< HEAD              ← your test_package_after_dependencies + test_dependencies_after_build
=======
                          ← master's test_lockfile_freshness
>>>>>>> origin/master
```

Keep both blocks and delete the three markers. `main()` merges cleanly on its own and already calls all four in the right order, so nothing else needs touching. I resolved it exactly that way locally and it builds with `ctest` 24/24.

### One heads-up on ordering

I have #78 open against the same region of `core/src/config.c` — a bounds fix for a stale `pkg_idx` writing past `EOS_MAX_PACKAGES` (AddressSanitizer-confirmed on master). Yours should go in first since it is the smaller change; I will rebase #78 on top of it rather than the other way round.

Ready to merge as soon as the rebase is up.

srpatcha (dismissed):

PR #75, opened 2026-08-28
fix(ota): stop treating the update's own hash as proof of authenticity

Kartikey1306:
## The architectural problem

The OTA service is the path that installs new firmware on a device. It has no notion of *who produced* an image.

```c
eos_ota_begin()  → memcpy(g_ota.expected_hash, source->expected_sha256, 32);
eos_ota_verify() → hash what we received, compare, set verified = 1
eos_ota_apply()  → if (!verified) return -1;  active_slot = update_slot;
```

Every input to that decision arrives **with the update**. An attacker who can substitute the image substitutes `expected_sha256` alongside it, and the check passes. This is an integrity check against transport corruption — a useful thing, but not what the API name, the `verified` flag, or `examples/secure-ota` claim it is.

What makes this architectural rather than a local bug is that **eos already has the other trust model, and the update path is the one place that does not use it.** `services/security` verifies a signature against a public key. `services/crypto` went further and adopted an explicit fail-closed policy — its stub verifiers now refuse to run unless a build defines `EOS_ALLOW_STUB_CRYPTO`, specifically so that a stub cannot forge a pass:

> *"So these now refuse to run unless a build explicitly opts in ... Anything else gets a hard failure instead of a forged pass."* — `services/crypto/src/rsa_ecc_sha512.c`

`services/ota` bypasses both. The tree has two trust models and firmware installation uses the weaker one.

## The change

Separate integrity from authenticity, and make the second one an explicit decision the platform has to make.

- The internal `verified` flag splits into **`integrity_ok`** and **`authenticated`**. `eos_ota_verify()` sets the first, and its doc comment now says plainly what it does and does not prove.

- **`eos_ota_set_authenticator()`** registers a platform callback. `eos_ota_apply()` calls it with the digest *the service computed over the bytes it actually received* — never the digest the update declared — plus any signature material carried on `eos_ota_source_t`, and installs nothing if it returns non-zero.

- With **no authenticator installed, `apply()` refuses** and prints why. A build that genuinely wants unsigned updates opts in with `EOS_ALLOW_UNSIGNED_OTA`, deliberately mirroring the `EOS_ALLOW_STUB_CRYPTO` convention already in the tree rather than inventing a second one.

### Two more defects in the same flow

- `eos_ota_verify()` set `slot_valid[update_slot] = 1`. A slot that had been downloaded and hashed but **never applied** therefore became a legal `eos_ota_rollback()` and `eos_ota_set_active_slot()` target — a half-installed image reachable as the device's fallback. Marking the slot moved to `apply()`.

- A second `eos_ota_verify()` call ran `eos_sha256_final()` on an already-finalized context. It now returns the first result.

### The example

`examples/secure-ota` is named for this, and its comments claimed `"verify SHA-256 signature"` and `"SHA-256 + signature"` while nothing anywhere checked a signature. It now registers an authenticator — one that **refuses**, because no signature verifier is linked into the build. Refusing to install is the correct behaviour for a device that cannot verify; returning 0 there would install anything.

## Second commit: making CI able to pass at all

While verifying the above I found two defects that fail `CI — eos` regardless of the code under test:

- **`.coveragerc` sets `fail_under = 100`**; the Python suite measures 95.65%. `Run Python tests` is `continue-on-error: false`, so the job failed on the coverage number with all 10 tests passing. Passing `--cov-fail-under=0` in CI is exactly how ebuild resolved the identical conflict, with the reasoning recorded in its own `.coveragerc`: the repo-wide ratchet belongs in `codecov.yml`. Both numbers are left alone — lowering `.coveragerc` to a real floor is a maintainer call. Same flag added to `python-tests.yml`, which has the same step.

- **The ARM job points `CMAKE_TOOLCHAIN_FILE` at `cmake/arm-cortex-m4.cmake`.** That directory holds only `cppcheck.cmake`, `deps.cmake` and `ebuild_layers.cmake` — the toolchain file lives in `toolchains/`. It also passes `-DBUILD_TESTS=OFF`, which is not this project's option name (`EOS_BUILD_TESTS`), so the flag was silently ignored.

## Verification

| Check | Result |
|---|---|
| `ctest --no-tests=error` (with ci.yml's exact flags: `-DEOS_BUILD_TESTS=ON -DEOS_PRODUCT=vbox_test`) | **24/24 pass** |
| `pytest tests/ --cov-fail-under=0` | 10 passed |

Four new tests in `tests/test_ota.c`:

- `apply()` refuses when no authenticator is installed
- `apply()` honours an authenticator that rejects, and does not switch slots
- the authenticator receives the digest the service computed, not the declared one
- an unapplied slot is not a rollback or `set_active_slot` target

The existing OTA tests are unchanged apart from installing an accepting authenticator where they previously relied on `apply()` succeeding unconditionally.

**Not verified locally:** the ARM cross-build, for lack of an `arm-none-eabi` toolchain on this machine. The toolchain file it now names exists and the option name matches `CMakeLists.txt`.

**Checks are currently `action_required`** — this is a fork PR, so a maintainer needs to approve the workflow runs before anything executes.

## Out of scope

`examples/secure-ota` does not compile standalone on `master` either — its `CMakeLists.txt` never defines `EOS_ENABLE_OTA`, so `eos/ota.h` compiles out entirely (21 errors before this branch, same after). Left alone deliberately; happy to fix it separately.


Kartikey1306:
Could a maintainer approve the workflow runs on this PR? As a fork contributor I can't — the approve endpoint returns `403 Must have admin rights to Repository`, so no check has executed yet.

Worth flagging before that approval gets spent: `CI — eos` could not have gone green on `master` either, for two reasons unrelated to this change. Both are fixed in the second commit here, so the run should come back clean rather than red for reasons that have nothing to do with the OTA work.

- `.coveragerc` sets `fail_under = 100` while the Python suite measures 95.65%. `Run Python tests` is `continue-on-error: false`, so the job failed on the coverage number with all 10 tests passing.
- The ARM job passes `-DCMAKE_TOOLCHAIN_FILE=cmake/arm-cortex-m4.cmake`, but `cmake/` holds only `cppcheck.cmake`, `deps.cmake` and `ebuild_layers.cmake` — the toolchain file is in `toolchains/`. It also passes `-DBUILD_TESTS=OFF`, which is not this project's option name (`EOS_BUILD_TESTS`), so the flag was silently ignored.

Locally, with the exact flags `ci.yml` configures (`-DEOS_BUILD_TESTS=ON -DEOS_PRODUCT=vbox_test`): **ctest 24/24 pass**, and `pytest tests/` 10 passed.

I haven't verified the ARM cross-build itself — no `arm-none-eabi` toolchain on this machine. The toolchain file it now names exists and the option name matches `CMakeLists.txt`.

srpatcha (approved):
Verified locally against current master: merges clean, builds with no new errors, `ctest` 26/26.

This is the right diagnosis and I want it in. Every input to the install decision arrived with the update, so `expected_sha256` was self-attested — an integrity check against transport corruption being presented by the API name, the `verified` flag and the example as proof of origin. Splitting `verified` into `integrity_ok` and `authenticated` is the fix that makes the distinction impossible to blur again.

Checked the two things that actually matter in the implementation:

- `apply()` passes `g_ota.actual_hash` to the authenticator, not `g_ota.expected_hash`. Handing an authenticator the digest the update declared would reproduce the original bug behind a signature-shaped API, so this is the load-bearing line.
- With no authenticator and no `EOS_ALLOW_UNSIGNED_OTA`, `apply()` returns -1 before touching `active_slot`. Fail-closed, and reusing the `EOS_ALLOW_STUB_CRYPTO` convention rather than inventing a second opt-in is the right instinct.

The `slot_valid[update_slot] = 1` in `verify()` is a good independent catch — a downloaded-and-hashed but never-applied image being a legal rollback target is a way to reach exactly the state the rest of this PR is preventing.

Making `examples/secure-ota` register an authenticator that *refuses* is the correct call. An example that returns 0 there teaches people to install anything, and the example was already claiming a signature check that did not exist.

One follow-up, not a blocker: the refusal uses `fprintf(stderr, ...)` while the tree has `EOS_ERROR` in `core/include/eos/log.h`, which is what `core/src/config.c` uses. On a target with no hosted stdio this is the one line in the change that will not build. Worth switching when convenient.

Not verified here: the ARM toolchain-path fix — no cross toolchain on this machine, so that half rests on CI.

PR #79, opened 2026-08-28
chore: untrack the .coverage database, and fix the last -DBUILD_TESTS=OFF

Kartikey1306:
Two loose ends from #75. Both are mine, and the first one is a mistake I made.

### `.coverage` is now tracked on master

`pytest --cov` writes a 52 KB SQLite database to the repo root. It is not in `.gitignore`, so a `git add -A` on my side swept it into #75 and it landed on `master`:

```
$ git cat-file -s $(git rev-parse origin/master:.coverage)
53248
$ git show origin/master:.coverage | head -c 16 | strings
SQLite format 3
```

It is local coverage state — machine-specific, changes on every run, and useless to anyone else. Worse, because it is not ignored, the next person to run `pytest --cov` sees a dirty tree and can commit their copy over mine.

Untracked here, and added to `.gitignore` along with `.coverage.*` (the per-process files `--cov` writes under parallel runs).

### The release job still passes `-DBUILD_TESTS=OFF`

#75 corrected the two occurrences in the test and ARM jobs. This third one, in `Build release artifacts`, landed just after the merge:

```yaml
cmake -B build/release -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_TESTS=OFF          # ← this project's option is EOS_BUILD_TESTS
```

`-DBUILD_TESTS=OFF` sets an unrelated cache variable, so the flag does nothing and the release build uses whatever `EOS_BUILD_TESTS` defaults to. Harmless today because that default is `OFF` — and it stops being harmless the moment the default changes or someone reads this line as documentation of intent.

### Verification

| Check | Result |
|---|---|
| `ctest --no-tests=error` | **26/26 pass** |
| `pytest --cov` afterwards | `.coverage` regenerates and `git status` stays clean |

No source changes.


Kartikey1306:
The red checks here are inherited from `master`, not caused by this PR — flagging so nobody spends time on them.

`EoS Build & Test` was already failing before this branch existed:

```
08-28T19:13  failure  2a37b3db   (#75)
08-28T19:10  failure  87ece28a   (#66)
08-28T19:05  failure  7b848118   (#48)
```

The cause is `tests/test_e2e.c` under `-Werror`:

```
tests/test_e2e.c:38:65: error: implicit declaration of function 'eos_tick_get'
tests/test_e2e.c:63:5:  error: this 'for' clause does not guard... [-Werror=misleading-indentation]
tests/test_e2e.c:63:59: error: comparison of integer expressions of different signedness
```

and `Python Unit Tests` fails earlier still, in `setup-python`'s cache step:

```
No file in /home/runner/work/eos/eos matched to [**/requirements.txt or **/pyproject.toml]
```

Neither is anything this PR touches — it deletes `.coverage`, edits `.gitignore`, and changes one CMake option name in `ci.yml`.

**#71 already fixes both** (`tests/test_e2e.c`, `kernel/include/eos/kernel.h`, `.github/workflows/build.yml`), so I have deliberately not duplicated it. Once #71 lands this should go green on a re-run; happy to rebase then if that's easier to confirm.

srpatcha (approved):

PR #84, opened 2026-08-29
crypto: refuse stub RSA/ECC signing unless EOS_ALLOW_STUB_CRYPTO

JoaoMorais03:
Contribution after reviewing the project: RSA/ECC **verify** already fail-closed unless `EOS_ALLOW_STUB_CRYPTO` is defined, but **sign** still minted fake signatures in production builds.

## Issue

`services/crypto/src/rsa_ecc_sha512.c` documents that the RSA/ECC routines are stubs, not PKCS#1/ECDSA. After the earlier verify-side fix, `eos_rsa_verify_sha256` / `eos_ecc_verify` refuse to run without the opt-in. `eos_rsa_sign_sha256` still wrote a PKCS#1-looking blob ending in the hash and returned 0; `eos_ecc_sign` still wrote a fake 64-byte r||s and returned 0. A signing pipeline could look like it worked while producing forgeries.

`tests/test_crypto_failclosed.c` already asserts that signing must refuse, and is listed among unregistered tests. It was not in `tests/CMakeLists.txt`.

## Approach

Call `EOS_STUB_CRYPTO_REFUSE` from both sign functions after the existing NULL checks, matching verify. Register `test_crypto_failclosed` against `eos_crypto` **without** `EOS_ALLOW_STUB_CRYPTO` so CI covers the production path. Existing `test_crypto_rsa` / `test_crypto_ecc` only `(void)` the sign/verify status, so they still pass.

## Testing

The fail-closed suite checks:
- the old RSA trailing-hash forgery is rejected
- RSA/ECC verify never succeed without the opt-in
- RSA/ECC sign refuse to mint fake signatures

I did not run CTest here (no local checkout). `test_crypto_rsa.c` / `test_crypto_ecc.c` were read to confirm they do not assert sign success.

## Limitations

Does not implement real RSA/ECDSA. Tests that want the old stub behavior must define `EOS_ALLOW_STUB_CRYPTO` (none of the currently registered tests do). Master currently has duplicate `test_crypto_aes` / `test_crypto_sha512` CMake targets, so a full test configure still fails until that is fixed independently; this PR only adds `test_crypto_failclosed`.


srpatcha (dismissed):
## Verified — this closes a real asymmetry

Checked `origin/master` before reviewing. The guards are on verify only:

```
50:int eos_rsa_sign_sha256(...)                      ← unguarded
67:int eos_rsa_verify_sha256(...)
69:    EOS_STUB_CRYPTO_REFUSE("eos_rsa_verify_sha256")
80:int eos_ecc_sign(...)                             ← unguarded
92:int eos_ecc_verify(...)
97:    EOS_STUB_CRYPTO_REFUSE("eos_ecc_verify")
```

So a build that correctly refuses to *verify* with stub crypto would still happily *sign* with it. Your reasoning for why that matters is the right one: a stub that mints a PKCS#1-looking blob or a fake `r||s` lets a signing pipeline appear to work end to end. The artifacts it produces are not signed by anything, and the failure surfaces at some later point — a real verifier rejecting them, or worse, a stub verifier accepting them — a long way from the code that caused it.

Making sign and verify symmetric is the correct fix, and reusing `EOS_ALLOW_STUB_CRYPTO` rather than inventing a second opt-in keeps one switch for one concept.

### The test is the durable part

`test_crypto_failclosed` deliberately does **not** define `EOS_ALLOW_STUB_CRYPTO`, so it exercises the production path rather than the test path. That is the distinction that makes it a guard: the existing crypto tests all opt in, so none of them would notice the refusal being removed.

The note to replace it with known-answer tests against published vectors when real verification lands is the right instruction to leave behind — a fail-closed test becomes actively misleading once the function works.

### One merge-order note

This conflicts with #82, which repairs `master`. Both touch `tests/CMakeLists.txt`: your new `test_crypto_failclosed` registration sits inside the duplicated `test_crypto_aes` / `test_crypto_sha512` block that #82 removes — the duplicate is why `master` currently fails to configure at all.

The resolution is to keep your `test_crypto_failclosed` lines and drop the duplicated aes/sha512 ones, which are already registered earlier in the file. I did exactly that locally and it is clean.

### Verification

Merged onto `origin/master` + #82, resolved as above: configure OK, **0 build errors, ctest 29/29**. Not verified against `master` alone, because `master` does not currently configure.

PR #93, opened 2026-08-30
fix(build): remove duplicate test targets from tests/CMakeLists.txt

prakhar7017:
## Problem

`tests/CMakeLists.txt` declared `test_crypto_aes` and `test_crypto_sha512` twice
each — once in a group block at lines 59–74, then again individually at lines
76–84. A repeated `add_executable` target name is a hard CMake error, not a
warning:

CMake Error at tests/CMakeLists.txt:77 (add_executable):
  add_executable cannot create target "test_crypto_aes" because another
  target with the same name already exists.  The existing target is an
  executable created in source directory "D:/eos/tests".

CMake Error at tests/CMakeLists.txt:79 (add_test):
  add_test given test NAME "test_crypto_aes" which already exists in this
  directory.

CMake Error at tests/CMakeLists.txt:82 (add_executable):
  add_executable cannot create target "test_crypto_sha512" because another
  target with the same name already exists.

-- Configuring incomplete, errors occurred!

Configuration aborted, so the exact command `CLAUDE.md` and `README.md` give for
building the suite —

```bash
cmake -B build/host -DCMAKE_BUILD_TYPE=Debug -DEOS_BUILD_TESTS=ON

— produced no build system at all. Not one test target could be generated,
built or run. This affects every developer following the documented setup.

Changes

- tests/CMakeLists.txt — removed the duplicate add_executable /
  target_link_libraries / add_test triples for test_crypto_aes and
  test_crypto_sha512. Kept the copy whose comment describes what the suite
  actually covers (AES-128/192/256 ECB + CBC, NIST FIPS-197 / SP 800-38A,
  SHA-512 against NIST vectors), and gave test_crypto_ecc and
  test_crypto_rsa the same one-line description so the block is consistent.
- CHANGELOG.md — entry under [Unreleased] / Fixed.

No target was added or removed, and no source file or link dependency changed.
The 28 registered test targets are exactly the same set as before, each now
declared once.

Testing

No new tests. This is a build-system repair — the evidence is that configuration
now completes.

┌───────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────┐
│                                 Check                                 │                  Result                  │
├───────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────┤
│ cmake -B build/pr1 -G Ninja -DCMAKE_BUILD_TYPE=Debug                  │ PASS — Configuring done / Generating     │
│ -DEOS_BUILD_TESTS=ON on this branch                                   │ done, no errors                          │
├───────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────┤
│ Same command on master                                                │ FAIL — 4 CMake errors, Configuring       │
│                                                                       │ incomplete                               │
└───────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────┘

Verified with CMake 4.4.3, Ninja 1.13.0, GCC 6.3.0 (MinGW) on Windows.

### CODE REVIEW

**Review by srpatcha — DISMISSED** [review]
The defect is real and I reproduced it on `master`:

```
$ git show origin/master:tests/CMakeLists.txt | grep -oE 'add_executable\([a-z_0-9]+' | sort | uniq -c | awk '$1>1'
      2 test_crypto_aes
      2 test_crypto_sha512
```

Two declarations each, and a duplicate `add_executable` name is a hard CMake
error, so `-DEOS_BUILD_TESTS=ON` cannot generate a single test target. The
CHANGELOG entry describes it accurately.

One thing you should know before this merges: **#82 fixes the same two
duplicates**, as part of repairing four breakages that landed together. It takes
`tests/CMakeLists.txt` from 30 `add_executable` lines to 28; so does this. The
two will conflict.

That does not make this PR wasted, because you made the better call on *which*
copy survives. #82 keeps the grouped comment:

```cmake
# --- test_crypto_aes / _ecc / _rsa / _sha512: were sitting in tests/ unused ---
```

which describes how the tests got there rather than what they cover. This keeps
the per-test descriptions:

```cmake
# --- test_crypto_aes: AES-128/192/256 ECB + CBC, NIST FIPS-197 / SP 800-38A ---
# --- test_crypto_ecc: ECC point arithmetic and ECDSA ---
# --- test_crypto_rsa: RSA modular exponentiation and verify ---
# --- test_crypto_sha512: SHA-512 against NIST vectors ---
```

Naming the standard each vector set comes from is worth more to the next reader
than a note about the file's history. I am folding your version into #82 with
attribution, so the outcome is yours even though the conflict resolves in that
PR's favour.

Approving on the merits. If #82 lands first this becomes empty and can close; if
this lands first, #82 rebases cleanly around it.

Worth flagging that this is not the only duplicate-target problem in the
platform. `eos` and `eBoot` *both* define `test_crypto` and `test_multicore`, and
since `ebuild` composes them with `add_subdirectory`, the integration build could
not configure either — same class of error, one repository apart. That one is
embeddedos-org/ebuild#85, fixed in embeddedos-org/eBoot#71.

PR #102, opened 2026-08-31
fix(hal): let the build system, not the preprocessor, pick the HAL backend

Kartikey1306:
Which backend source CMake compiles, and which registrar `eos_hal_init()` calls, are **the same decision made in two places with two different conditions**.

```cmake
# CMakeLists.txt
if(EOS_PLATFORM STREQUAL "linux" OR NOT CMAKE_CROSSCOMPILING)
    target_sources(eos_hal PRIVATE hal/src/hal_linux.c)
else()
    target_sources(eos_hal PRIVATE hal/src/hal_rtos.c)
```

```c
/* hal/src/hal_common.c */
#ifdef __linux__
    eos_hal_linux_register();
#else
    eos_hal_rtos_register();
```

They agree on Linux and on a cross build, and disagree on **every other host**. A macOS host build compiles `hal_linux.c` and then references a symbol from the file it did not compile:

```
  "_eos_hal_rtos_register", referenced from: _eos_hal_init in libeos_hal.a
```

## The change

CMake sets `EOS_HAL_BACKEND_HOST` or `EOS_HAL_BACKEND_RTOS` right next to the `target_sources()` call that makes the choice, and `hal_common.c` keys off that. `__linux__` stays as a fallback for anyone compiling these files without CMake.

Two things had to move with it:

- **`hal.h` declared the registrars under the same `#ifdef`**, so `hal_common.c` could not even *name* the one CMake had compiled. Both are declared unconditionally now — exactly one is still defined per build, and an unreferenced declaration costs nothing. The header's own comment already argued for this: *"an undeclared registrar is one nobody can find."*
- **`hal_linux.c` wraps its whole body in `#ifdef __linux__`**, so on macOS it compiled to an empty translation unit and the build had no backend at all. It is the *host* backend rather than the *Linux* one — the only platform calls in it are `clock_gettime()` and `nanosleep()`, both POSIX. `hal_rtos.c` takes the mirror condition.

## Linux is unchanged

`EOS_HAL_BACKEND_HOST` and `__linux__` are both defined there, so the same registrar is selected as before. Cross builds get `EOS_HAL_BACKEND_RTOS` and are likewise unchanged. `master` CI is green on Linux and stays that way — this is not a fix *for* CI, it is a fix for an inconsistency CI cannot see.

## Scope — stated plainly

This does **not** make macOS a supported host on its own.

| | before | after |
|---|---|---|
| `eos_hal` link | fails | links |
| `ctest` on macOS | 0/28 | **19/28** |

The remaining 9 failures are `services/compat/src/os_adapter_posix.c` calling `pthread_mutex_timedlock()` and `sem_timedwait()`, neither of which macOS provides. That is separate work in a different subsystem and I have not attempted it here — flagging it rather than half-doing it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,10 @@ on:
 # cancel-in-progress because a superseded commit's result is not wanted.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # Cancel superseded runs on PR branches, but never on master: cancelling
+  # master's runs on every merge is how a hanging test stayed invisible --
+  # no master run had been allowed to finish since 08-30.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
 
 env:
   BUILD_TYPE: Release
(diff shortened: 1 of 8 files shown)


Kartikey1306:
Closing — this landed. Merging current master into the branch leaves an empty diff: `EOS_HAL_HOSTED` is on master, comment and all, carried in when #108/#109/#110 merged from on top of this.

Thank you for taking the stack. Everything that was blocked behind it is now unblocked, and #104/#105/#106 are rebased onto master directly rather than onto this.

PR #113, opened 2026-09-01
fix(net): guard net_posix.c, which broke every non-test build

Kartikey1306:
> Stacked on #112. Review the last commit.

## The default build does not compile

```
$ cmake -S . -B build && cmake --build build
net/src/net_posix.c:43:30: error: unknown type name 'eos_net_addr_t'
net/src/net_posix.c:73:1:  error: unknown type name 'eos_socket_t'
... 30 more
```

This is what the library job in `CI — eos` runs, and it is failing on master for this reason.

## Why

`eos/net.h` puts every eNet declaration behind `#if EOS_ENABLE_NET`. `CMakeLists.txt` defines that flag only under:

```cmake
if(EOS_BUILD_TESTS AND NOT EOS_PRODUCT)
    add_compile_definitions(... EOS_ENABLE_NET=1 ...)
```

but `net/src/net_posix.c` is in `eos_net`'s source list **unconditionally**. Built without the feature, the header contributes no types and every definition in the file fails.

That is also why it stays invisible locally: `-DEOS_BUILD_TESTS=ON` is what anyone running the suite passes, and it happens to be the one configuration that defines the flag.

## Fix

Guard the body the way the header is guarded, so the translation unit compiles away when the feature is off — what a disabled feature is supposed to do — rather than failing to compile.

## Verified in both configurations

```
$ cmake -S . -B build && cmake --build build          # was: 32 errors
default build OK

$ cmake -S . -B build -DEOS_BUILD_TESTS=ON && ctest
100% tests passed, 34 of 34

$ nm libeos_net.a | grep -c eos_net_socket
1                                                      # still linked when the feature is on
```

### DIFF
--- a/net/src/net_posix.c
+++ b/net/src/net_posix.c
@@ -25,6 +25,15 @@
 
 #include "eos/net.h"
 
+/* eos/net.h puts every eNet declaration behind EOS_ENABLE_NET, but this
+ * file is in eos_net's source list unconditionally. Built without the
+ * feature -- which is the default configuration, since CMakeLists.txt only
+ * defines EOS_ENABLE_NET for the test build -- the header contributed no
+ * types and every definition below failed with "unknown type name
+ * eos_socket_t". Guarding the body compiles it away instead, which is what
+ * a disabled feature is supposed to do. */
+#if EOS_ENABLE_NET
+
 #include <errno.h>
 #include <netdb.h>
 #include <netinet/in.h>
@@ -203,3 +212,5 @@ int eos_net_resolve(const char *hostname, uint32_t *ip)
     freeaddrinfo(res);
     return 0;
 }
+
+#endif /* EOS_ENABLE_NET */
--- a/tests/test_net.c
+++ b/tests/test_net.c
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: MIT
+#include <assert.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>


Kartikey1306:
Closing this in favour of **#116**, which covers the same two failures. Both PRs came from this account within a few hours — my fault for not checking first, and I'd rather leave one PR on the bug than two.

**`net_posix.c` — functionally identical.** Both add `#if EOS_ENABLE_NET` around the body; only the placement relative to the system includes and the comment wording differ. Nothing to choose between them.

**`test_net.c` — this is why #116 rather than this one.** This PR adds `#include <assert.h>`, which does fix the link error. #116 converts the ten call sites to the `CHECK()` macro the file already defines, and the file's own comment explains why that distinction matters:

> Not CHECK(): CI builds Release, and NDEBUG deletes CHECK() together with the expression inside it. Half these checks wrap the call under test — `CHECK(eos_net_connect(...) == 0)` stopped connecting at all, the echo thread blocked in `accept()` forever, and `pthread_join()` hung the suite. A check macro must always evaluate its expression.

Adding the include makes those ten work only while `add_compile_options(-UNDEBUG)` stays in `tests/CMakeLists.txt`. If that line is ever moved, scoped differently, or a target is built outside it, `assert()` goes back to being deleted along with `assert(eos_net_connect(...) == 0)` — and the suite goes back to hanging in `accept()` rather than failing. That is the exact defect the `CHECK()` macro was added for, and four of the ten sites here are `eos_net_connect` / `eos_net_send` / `eos_net_recv` calls.

#116 also carries the verification I could not do from this branch: the default-configuration build reproduced against unpatched `origin/master` as a negative control, plus `medical`/`aerospace` profiles and `ctest 34/34`.

Nothing in this PR is lost.

PR #116, opened 2026-09-01
chore(tests): drop the unused assert.h include from test_net.c

Kartikey1306:
`CI — eos` has been red on master since 9861a3b. Two independent breakages, both fixed here.

## 1. The default configuration does not compile

```
net/src/net_posix.c:43:30: error: unknown type name 'eos_net_addr_t'
net/src/net_posix.c:73:29: error: unknown type name 'eos_net_proto_t'
```

`net/include/eos/net.h` wraps everything it declares in `#if EOS_ENABLE_NET`. Nothing sets that flag unless a product profile selects it or `EOS_BUILD_TESTS AND NOT EOS_PRODUCT` turns it on. But CMake adds the source unconditionally:

```cmake
if(NOT CMAKE_CROSSCOMPILING)
    target_sources(eos_net PRIVATE net/src/net_posix.c)
```

`EOS_ENABLE_NET` never enters that decision, so on any host build without tests or a product the header expands to nothing while `net_posix.c` still defines the API against it.

`net.c` already carries `#if EOS_ENABLE_NET` around its body and survives untouched; the POSIX mapping was the half that never got the guard. Added, so the two translation units appear and disappear together.

Worth noting: the `Default configuration` job's own comment says it exists because *"that gap let a translation unit that only compiles under EOS_ENABLE_NET reach master."* It was added for exactly this class, and it caught exactly this.

## 2. `test_net` does not link

```
undefined reference to `assert'
```

`test_net.c` defines `CHECK()` and explains why in a comment above it — CI builds Release, `NDEBUG` deletes `assert()` *together with the expression inside it*, and half these checks wrap the call under test. `assert(eos_net_connect(...) == 0)` stopped connecting at all, the echo thread blocked in `accept()`, and the suite hung.

Ten call sites in the `EOS_NET_POSIX` section still used bare `assert()` — including that exact `eos_net_connect` line. They were dormant while `assert()` was a no-op. #104 added `-UNDEBUG` to the test targets, which is correct and is what makes the other 738 assertions in this suite real; it also made these ten live. The file never includes `<assert.h>`, so `assert` resolved to an implicit function declaration and went undefined at link.

Converted to `CHECK()` rather than adding the include: the include would make them real assertions again in Debug and silently-deleted ones in Release, which is the defect the macro was written to prevent.

## Validation

Verified locally on macOS, all three configurations CI builds:

| | |
|---|---|
| `cmake -B build/host -DCMAKE_BUILD_TYPE=Release` | builds — **fails on master** |
| `-DEOS_PRODUCT=medical`, `-DEOS_PRODUCT=aerospace` | both build |
| `-DEOS_BUILD_TESTS=ON -DEOS_PRODUCT=vbox_test` + ctest | **34/34 passed**, `test_net` among them |

**Negative control:** the same default-configuration build against unpatched `origin/master` in a clean worktree reproduces the `unknown type name` errors verbatim, so the fix is doing the work and not the build directory.

### DIFF
--- a/.github/workflows/eos-simulation.yml
+++ b/.github/workflows/eos-simulation.yml
@@ -9,6 +9,15 @@ on:
 
 env:
   EOS_VERSION: "0.6.0"
+  # The kernel job compiles eBoot and checks out ebuild. Those used to float on
+  # `ref: master`, so the moment either repo's master stopped compiling, every
+  # open eos PR went red for a reason no eos change caused and no eos change
+  # could fix (2026-09-03: eBoot master's eos_image.h asserts a member that no
+  # longer exists, and its ed25519_verify.c has a redefined helper -- eBoot #94
+  # repairs it). A dependency is pinned and bumped deliberately; when eBoot #94
+  # lands, bump EBOOT_COMMIT to that merge and CI will say whether it holds.
+  EBOOT_COMMIT: "a172a6d6e1e66877413ed546401a65ee4f4f90db"
+  EBUILD_COMMIT: "e5d8052f3e2cfdcb1c91bab8300087aa07e02679"
   CROSS_COMPILE: "aarch64-linux-gnu-"
   QEMU_MACHINE: "virt"
   QEMU_CPU: "cortex-a57"
@@ -29,14 +38,14 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: embeddedos-org/eBoot
-          ref: master
+          ref: ${{ env.EBOOT_COMMIT }}
           path: eBoot
 
       - name: Checkout ebuild
         uses: actions/checkout@v4
         with:
           repository: embeddedos-org/ebuild
-          ref: master
+          ref: ${{ env.EBUILD_COMMIT }}
           path: ebuild
 
       # embeddedos-org/eFab does not exist -- the checkout 404s and takes this
@@ -276,7 +285,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: embeddedos-org/ebuild
-          ref: master
+          ref: ${{ env.EBUILD_COMMIT }}
 
       # Same missing repository as in build-kernel.
       # - name: Checkout eFab
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -90,6 +90,11 @@ add_executable(test_crypto_rsa test_crypto_rsa.c)
 target_link_libraries(test_crypto_rsa PRIVATE eos_crypto)
 add_test(NAME test_crypto_rsa COMMAND test_crypto_rsa)
 
+# --- test_crypto_ed25519_loworder: reject keys outside the prime-order subgroup ---
+add_executable(test_crypto_ed25519_loworder test_crypto_ed25519_loworder.c)
+target_link_libraries(test_crypto_ed25519_loworder PRIVATE eos_crypto)
+add_test(NAME test_crypto_ed25519_loworder COMMAND test_crypto_ed25519_loworder)
+
 # --- test_crypto_sha512: SHA-512 against NIST vectors ---
 add_executable(test_crypto_sha512 test_crypto_sha512.c)
 target_link_libraries(test_crypto_sha512 PRIVATE eos_crypto)
--- a/tests/test_net.c
+++ b/tests/test_net.c
@@ -2,7 +2,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <assert.h>
 #define EOS_ENABLE_NET 1
 #include "eos/eos_config.h"
 #include "eos/net.h"


Kartikey1306:
⚠️ **#114 and #116 both add the `#if EOS_ENABLE_NET` guard to `net/src/net_posix.c`, and merging both breaks master — with no merge conflict to warn you.**

Both PRs came from this account. #116 fixes the guard plus the `test_net` link error; #114 fixes the same guard plus adds a CMake test-registration check. The overlap is `net/src/net_posix.c` and `tests/test_net.c`.

Git merges them silently, because each adds its guard at a different point in the include block:

```
$ git merge fork/test/register-orphaned-suites   # onto #116
Auto-merging net/src/net_posix.c
Auto-merging tests/test_net.c
   (no conflict)

$ grep -c '^#if EOS_ENABLE_NET'            net/src/net_posix.c   ->  2
$ grep -c '^#endif /\* EOS_ENABLE_NET \*/' net/src/net_posix.c   ->  1

$ cmake -B build/host -DCMAKE_BUILD_TYPE=Release && cmake --build build/host
net/src/net_posix.c:35:2: error: unterminated conditional directive
```

Two opening guards, one closing. The default-configuration build fails exactly the way it does on master today, so merging both would put back the breakage both PRs exist to remove.

**Suggested resolution:** take **#116** for the `net_posix.c` guard and the `test_net.c` fix, and drop those two files from **#114**, leaving it the `tests/CMakeLists.txt` registration guard and `tests/unit/test_cmake_test_registration.py` — which is the part of #114 that is genuinely its own and worth having. In either order, the second one merged then needs a rebase rather than a straight merge.

Related: I closed #113 in favour of #116 for the same overlap.

srpatcha (commented):
## Review — eos#116 "fix: repair master — the default configuration and the test link"
head: d58bac4439e6a14433336ba5a249108ac6b1528f  author: Kartikey1306  ci: pass

**Verdict:** Both repairs have already landed on `master` by another route. Against
current `master` this PR reduces to a single line — removing an unused
`#include <assert.h>` from `tests/test_net.c` — and the title's claim to repair the
default configuration no longer holds. The remaining line is correct but does not need
this PR.

### Findings

|---|---|---|---|---|
| 1 | Medium | PR title and body | "repair master — the default configuration and the test link" is no longer true at this head. `net/src/net_posix.c` is byte-identical to `origin/master`, and `master`'s default configuration builds clean. Per the brief, an unsupported claim in a PR body is itself the finding; here it would also mislead a maintainer into merging this ahead of work that actually is blocking. | Close the PR, or retitle it to what it now does — `chore(tests): drop the unused assert.h include from test_net.c` — and rewrite the body. |
| 2 | Low | `tests/test_net.c:5` | The one surviving change removes `#include <assert.h>`. That include was added to `master` one commit ago by #124, whose message reads "add the missing assert.h include that broke master". `tests/test_net.c` contains zero `assert()` calls at #124's parent and at #124 itself — the file uses its own `CHECK()` macro (`test_net.c:18`, 22 call sites). So the include is genuinely unused and removing it is right, but it reverts a commit whose stated reason does not match the file. Worth resolving with #124's author before it ping-pongs. | Land the removal, and correct the record on #124 so the next person does not re-add it. |
| 3 | Low | `net/src/net_posix.c:44` (on `master`, unchanged by this PR) | The `#if EOS_ENABLE_NET` guard fixes the symptom; the cause is that `CMakeLists.txt:201` adds `net_posix.c` to `eos_net` on `if(NOT CMAKE_CROSSCOMPILING)` — a platform condition — while whether the file has any contents is decided by `EOS_ENABLE_NET`, a C macro defaulting to `0` (`include/eos/eos_config.h:108`) and set to `1` only under `if(EOS_BUILD_TESTS AND NOT EOS_PRODUCT)` (`CMakeLists.txt:48-51`). The two conditions have no common source of truth, so they stay in agreement only by hand. Any other source added on a platform condition but bodied on a feature flag reproduces this exactly. | Promote `EOS_ENABLE_NET` to a real CMake `option()` that both drives `target_compile_definitions` and gates `target_sources`, so the file and its contents appear and disappear together by construction. Tracked as a design-doc proposal rather than a change to ask of this PR. |
| 4 | Low | `net/src/net.c:15` | Pre-existing warning surfaced while building this head: `warning: 'net_initialized' defined but not used [-Wunused-variable]`. With `EOS_ENABLE_NET=0` the variable's only users are compiled out. | Move the definition inside the `#if EOS_ENABLE_NET` block in `net.c`. |

Evidence, against `origin/master` at `feee272`:

```
$ git diff origin/master refs/pull/116/head -- net/src/net_posix.c
                                        # no output — identical
$ git diff origin/master refs/pull/116/head -- tests/test_net.c
-#include <assert.h>
                                        # the whole remaining diff

$ grep -c '^#if EOS_ENABLE_NET' net/src/net_posix.c      # on master
1
$ cmake -B build/def && cmake --build build/def          # on master, default config
# exit 0
```

`#124` and the include it added:

```
$ git show origin/master --stat
 tests/test_net.c | 1 +
$ git show feee272^:tests/test_net.c | grep -c 'assert('
0
$ git show feee272^:tests/test_net.c | grep -n 'define CHECK'
17:#define CHECK(cond) do { \
```

And the removal is safe:

```
$ cmake -B build/host -DEOS_BUILD_TESTS=ON && cmake --build build/host --target test_net
# exit 0, no error
$ ./build/host/tests/test_net
# exit 0
```

The `assert` → `CHECK` conversion this PR was credited with is the better of the two
available fixes — `CHECK()` survives `-DNDEBUG`, which is the whole point of #104 — and
it is already on `master`. Nothing here argues against the approach; it argues that the
work is done.

### Architecture conformance

Conforms. Master design §21 places `eos` in Tier 1 - Foundation; `net/` is the eNet
connectivity module named in §5 under Core Services → Connectivity, and it lives in the
owning repo rather than a separate one, which is correct under §21.1. No include, link
or manifest edge points up a tier: `net_posix.c` reaches only `<netdb.h>`,
`<sys/socket.h>` and `eos/net.h`. §5.1 holds.

Finding 3 is where the master design is silent rather than wrong, and I have appended a
proposal for it under `.ai/autoreview/proposals/2026-09.md`. §9.2 requires "one source
of truth for CLI, VS Code and EoStudio" and §10.1 gives components a Capabilities field,
but no section governs the relationship between build-system feature selection and the
C-level `EOS_ENABLE_*` macros that decide whether the selected code has any contents.
That gap is what broke the default configuration here.

### Proposed changes

1. Confirm with #124's author that `tests/test_net.c` needs no `assert.h` — the
   commands above settle it — then either close this PR or reduce it to that one line
   with an accurate title.
2. Keep #114 and #115, which are stacked on this commit, moving. Because both carry
   `d58bac4`, merging either one first brings this content along; closing this PR does
   not lose anything.
3. Finding 4 is a one-line move and can ride with whichever PR next touches `net/`.

Do not merge this alongside a second PR that adds the same guard. That hazard was real
when #114 was a sibling; it is now resolved, because #114 is stacked on this commit and
carries exactly one `#if EOS_ENABLE_NET` / one `#endif` (verified below), so no
double-guard can arise from this pair:

```
$ git show refs/pull/114/head:net/src/net_posix.c | grep -c '^#if EOS_ENABLE_NET'
1
$ git show refs/pull/114/head:net/src/net_posix.c | grep -c '^#endif /\* EOS_ENABLE_NET \*/'
1
```

### Not checked

- I did not establish what actually motivated #124. I established only that
  `tests/test_net.c` has no `assert()` call and builds without the include; if #124 was
  fixing a different symptom, that reason is not visible in the file it changed.
- I did not build any `EOS_PRODUCT` profile. `EOS_ENABLE_NET` is set to `1` only under
  `EOS_BUILD_TESTS AND NOT EOS_PRODUCT`, so whether a product build gets eNet at all
  depends on profile files I did not inspect. Finding 3 describes the mechanism, not the
  consequence for products.
- I did not run the cross-compile configurations. `net_posix.c` is excluded under
  `CMAKE_CROSSCOMPILING`, so the guard should be inert there, but I verified only the
  host default and host test builds.
- CI is green at this head; I read the check list, not the job logs.
- I did not review `d58bac4` a second time in the #114 and #115 reports. It is the
  shared base of all three.


---
*Automated architecture review of `d58bac4439e6` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (commented):
## Review — eos#116 "chore(tests): drop the unused assert.h include from test_net.c"
head: fafec15061c9cfcccd967989239e032233eb2746  author: Kartikey1306  ci: 2 required checks failing (pre-existing — see below)

**Verdict:** The diff is correct, minimal and worth landing: it removes an include with no remaining users, and it registers `test_crypto_ed25519_loworder`, a real security suite that has been sitting in `tests/` unbuilt. Verified 39/39 against master's 38/38, and the extra test is the new one. The problem is the PR body, which describes an entirely different change from the one in the diff.

### Findings

|---|---|---|---|---|
| 1 | Medium | PR body | The body describes work that is not in this diff and is already on master. It says this PR adds `#if EOS_ENABLE_NET` to `net/src/net_posix.c` — that guard is already there on `origin/master` at `net/src/net_posix.c:44` and `:216`, and this diff does not touch the file. It says ten bare `assert()` calls in the `EOS_NET_POSIX` section are converted to `CHECK()` — `tests/test_net.c` on this head contains **zero** occurrences of the string `assert` at all, and the diff's only change to it is deleting the `#include`. The "Validation" table then reports `34/34 passed` for a tree that no longer exists; this head runs 39. The title is accurate; the body is a stale artifact of a rebase. Per the brief, evidence attached to the wrong change is worse than no evidence, because it reads as verification. | Rewrite the body to the two things the diff does: delete a now-unused include that `#124` added, and register the orphaned `test_crypto_ed25519_loworder` suite. Re-run and paste `39/39`. |
| 2 | Low | existing comment, 2026-09-01T15:30:58Z | The pinned warning — "**#114 and #116 both add the `#if EOS_ENABLE_NET` guard to `net/src/net_posix.c`, and merging both breaks master — with no merge conflict to warn you**" — no longer applies to either PR. Neither head touches `net/src/net_posix.c`: this PR's changed files are `tests/CMakeLists.txt` and `tests/test_net.c`; #114's are `tests/CMakeLists.txt` and `tests/unit/test_cmake_test_registration.py`. The `unterminated conditional directive` scenario it demonstrates cannot occur from these two heads. A maintainer reading it will hold both PRs for a hazard that was fixed by the rebase. | Edit or delete the comment. It was a good catch when written; leaving it is now the misleading part. |
| 3 | Low | CI | Two required checks are red on this head: **`Cross-compile ARM64 kernel`** and **`Full-stack integration summary`**. They are not caused by this PR — the identical pair fails on all four open `eos` heads in this run, including #117, whose entire diff is seven lines of `.gitignore` plus deletion of committed build artifacts. So this is `master` being red, and this PR inheriting it. Naming it because the brief requires failing required checks to be named, not because it is yours. | Nothing to do here. It does mean this PR cannot be merged on a green tree until master's ARM64 cross-compile is fixed, which is worth knowing before you wait on it. |

### Architecture conformance

Conforms. `eos` is Tier 1 Foundation (§21) and both files are under `tests/`, which is the directory `.ai/architect.md` names in the reference layout for exactly this. No `#include`, link line or `target_link_libraries` entry points up a tier: the new registration links `test_crypto_ed25519_loworder` against `eos_crypto`, which is a core service inside the same repository, and adds no cross-repo edge. §5.1's dependency direction is untouched.

The registration is also the more valuable half of the diff, and it lands in the right place. `tests/test_crypto_ed25519_loworder.c` has been in the tree, unreferenced by `tests/CMakeLists.txt`, so a suite whose subject is rejecting keys that let an attacker forge a signature has never been compiled or run. That is the same defect class `.ai/security.md` calls out — "a security suite that silently collected nothing is a failed check reported as green" — one level up: not a suite that collected nothing, a suite that was never built. §28's evidence table puts "Implemented" at "code and functional tests"; an unbuilt test file does not meet it, and this diff is what does.

### Proposed changes

1. Replace the PR body. The two sentences it needs are the ones in the title and the registration, plus the current `ctest` output.
2. Update or remove the 2026-09-01 cross-PR warning comment.
3. Nothing in the code. I would land the diff as it stands.

One thing worth doing in a follow-up, not here: `test_crypto_ed25519_loworder.c` was not the only orphan — this diff registers one file, and #114 adds `tests/unit/test_cmake_test_registration.py` to catch the class. Those two belong together, and #114's guard is what stops the next one going unnoticed.

### Not checked

- **Not run:** the two failing required checks. `Cross-compile ARM64 kernel` and `Full-stack integration summary` need a cross toolchain and the integration harness; I did not reproduce either, so I can state only that they fail on all four `eos` heads in this run and are therefore not attributable to this diff. I did **not** confirm the root cause on master.
- **Not verified:** the body's claim about macOS. All my builds are Linux/GCC; the body says "Verified locally on macOS", and nothing here speaks to that platform.
- **Not verified:** that `test_crypto_ed25519_loworder` is *correct*, only that it builds, runs and passes its own five cases. I did not check the low-order encodings it uses against the curve, nor confirm that `eos_crypto` rejects them for the right reason rather than incidentally.
- **Not checked:** whether other test files under `tests/` are also unregistered. That is #114's subject and I looked at this one file only.
- **Not checked:** the `-DEOS_PRODUCT=medical` and `-DEOS_PRODUCT=aerospace` configurations named in the body. I built the default host configuration and `vbox_test`.

### Evidence

Read-only `git archive` extractions of `origin/master` (`feee272`) and this head; the `eos` checkout was clean and untouched.

```
origin/master  cmake -B build/test -DCMAKE_BUILD_TYPE=Release -DEOS_BUILD_TESTS=ON \
                     -DEOS_PRODUCT=vbox_test && cmake --build            -> OK
               ctest --no-tests=error                                    -> 38/38 passed

this head      same configuration                                        -> OK
               ctest --no-tests=error                                    -> 39/39 passed
               cmake -B build/host -DCMAKE_BUILD_TYPE=Release            -> OK  (default config)

the +1:        ctest -R test_crypto_ed25519_loworder                     -> Passed
               ./tests/test_crypto_ed25519_loworder
                   test_all_zero_key_is_rejected                  [PASS]
                   test_identity_key_is_rejected                  [PASS]
                   test_every_low_order_key_is_rejected           [PASS]
                   test_rfc8032_vector_still_verifies             [PASS]
                   test_rfc8032_vector_rejects_a_tampered_message [PASS]
                   5/5 tests passed

safety of the deletion:
    tests/test_net.c @ this head : grep -c assert  -> 0
    tests/test_net.c @ master     : grep -n assert -> 5:#include <assert.h>   (the include only)

finding 1, against the body:
    net/src/net_posix.c @ master : line  44  #if EOS_ENABLE_NET
                                   line 216  #endif /* EOS_ENABLE_NET */   -> guard already merged
    net/src/net_posix.c          : not in this PR's changed-file list at all
    tests/test_crypto_ed25519_loworder.c exists on master, and
    tests/CMakeLists.txt @ master : grep -n ed25519 -> no match            -> orphaned until now
```


---
*Automated architecture review of `fafec15061c9` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (commented):
## Review — eos#116 "chore(tests): drop the unused assert.h include from test_net.c"
head: 2a32e3a919c161e62f8323f2138a268f6828272b  author: Kartikey1306  ci: 20 pass, 3 pending, 0 failing (see the note on `Cross-compile ARM Cortex-M4` below)

**Verdict:** The two changes that match the title are still correct and still worth landing — verified at this head: default host configuration builds, and `ctest --no-tests=error` gives **39/39** against master's 38, the extra being the newly registered `test_crypto_ed25519_loworder`. `tests/test_net.c` contains zero occurrences of `assert` at this head, so deleting the include is safe. What changed since `fafec150` is that the PR grew a third change which is a byte-identical copy of another open PR's, and the body still describes none of the three.

### Findings
|---|---|---|---|---|
| 1 | Medium | `.github/workflows/eos-simulation.yml:11-19`, `:41`, `:48`, `:288` | `2a32e3a` ("ci: pin the eBoot and ebuild checkouts instead of floating on master") is the same change as **#135**, which carries the same commit title and is open and under review. Verified: `git diff <this head> <#135 head> -- .github/workflows/eos-simulation.yml` is **empty** — the `env:` block, the comment, and all three `ref:` substitutions are identical, byte for byte. #135 has since added `f34edb3` ("ci: make the pin visible when it goes stale"), which this copy does not have. Identical hunks merge without conflict, so nothing breaks; the cost is that a cross-repository CI-dependency change is now sitting in a PR titled "drop the unused assert.h include", where a maintainer reviewing the title will not look for it and a maintainer reviewing #135 will not know it is duplicated. That is the same overlap hazard you documented yourself for #114/#116 on 2026-09-01. | Drop `2a32e3a` from this branch. #135 is the right home: it is scoped to the pin, it is the head that has the staleness guard, and it has already been reviewed twice. If the pin is needed here to get a green run, say so on the thread rather than carrying the commit. The pin's substance — the missing staleness signal, and the ten-repo middleware matrix still on `ref: master` at `:248` — is already reviewed on #135 and is not re-raised here. |
| 2 | Medium | PR body | Unresolved from the `fafec150` review, and now wrong about three things instead of two. The body says this PR adds `#if EOS_ENABLE_NET` to `net/src/net_posix.c` — the guard is already on `origin/master` (`feee272`) and this diff does not touch the file. It says ten bare `assert()` calls are converted to `CHECK()` — `tests/test_net.c` at this head has no `assert` in it at all and the diff's only change to it is the deleted `#include`. Its "Validation" table reports `34/34`; this head runs **39/39**. And it says nothing whatever about the `eos-simulation.yml` pin, which is now the largest part of the diff. Per the brief, a PR body asserting "verified" against a tree that no longer exists is itself the finding, because it reads as evidence for the change in front of you. | Rewrite the body to the diff: delete an include that `#124` made redundant, register the orphaned `test_crypto_ed25519_loworder` suite, and (if `2a32e3a` stays) the pin. Paste the current `39/39`. |
| 3 | Low | existing comment, 2026-09-01T15:30:58Z | Also unresolved from the previous review. The pinned warning — "#114 and #116 both add the `#if EOS_ENABLE_NET` guard to `net/src/net_posix.c`, and merging both breaks master" — is true of neither head any more: this PR's files are `.github/workflows/eos-simulation.yml`, `tests/CMakeLists.txt` and `tests/test_net.c`; #114's are `tests/CMakeLists.txt` and `tests/unit/test_cmake_test_registration.py`. Neither touches `net/src/net_posix.c`, so the `unterminated conditional directive` it demonstrates cannot happen. It was a good catch when written and is now the thing holding both PRs. | Edit or delete it. The real overlap between these two heads today is `tests/CMakeLists.txt`, which is a different and much smaller problem — see #114's review. |

### Architecture conformance
Conforms. `eos` is Tier 1 — Foundation (§21) and all three files are infrastructure: `tests/` is the directory `.ai/architect.md` names in the reference layout, and `.github/workflows/` is Infrastructure per §21. No `#include`, link line or `target_link_libraries` entry points up a tier — the new registration links `test_crypto_ed25519_loworder` against `eos_crypto`, a core service in the same repository, adding no cross-repo edge, so §5.1's runtime direction is untouched.

The pin in `2a32e3a` is a **CI-time** dependency from `eos` onto `eBoot` and `ebuild`, both Tier 1, so it points sideways rather than up. §5.1 governs runtime dependencies and is silent on this class; that gap and this exact pin are already recorded in `.ai/autoreview/proposals/2026-09.md` (the 2026-09-03 entry on build- and CI-time dependencies, and the 2026-09-04 addendum triggered by #135, which analyses this same SHA). Nothing new to propose from this PR. I did confirm the pin is a real, reachable commit: `a172a6d` is `fix(ed25519): reject public keys outside the prime-order subgroup (#86)` and is an ancestor of eBoot's `origin/master`; `e5d8052` is `ebuild`'s current master merge.

The registration remains the more valuable half of the diff and lands in the right place. `tests/test_crypto_ed25519_loworder.c` was in the tree unreferenced by `tests/CMakeLists.txt`, so a suite whose subject is rejecting keys that let an attacker forge a signature had never been compiled. That is the defect class `.ai/security.md` names — "a security suite that silently collected nothing is a failed check reported as green" — one level up, and §28's "Implemented = code and functional tests" is not met by a test file nothing builds.

### Proposed changes
1. Drop `2a32e3a`; let #135 carry the pin.
2. Rewrite the body against the actual diff and paste `39/39`.
3. Edit or delete the 2026-09-01 cross-PR comment.
4. Nothing in the C or CMake. I would land the remaining two-file diff as it stands.

### Not checked
- **`Analyze (C/C++)`, `Cross-compile ARM Cortex-M4` and `QEMU ARM64 simulation (13 TCs)` were still pending when this bundle was built**, so I cannot report their outcome. Nothing was failing. Worth recording that `Cross-compile ARM64 kernel` — red on this PR at `fafec150` and on every other open `eos` head in the previous run — is **green** at this head, which is the pin doing its job.
- **`Cross-compile ARM Cortex-M4` appears twice in `checks.txt`, once `pass` and once `pending`.** Two workflows declare the same check name: `.github/workflows/build.yml:74` and `.github/workflows/ci.yml:117`. Not introduced by this PR, and not a finding against it, but it means "is that check green" has no single answer on any `eos` PR — the same defect ebuild#104 is fixing in `ebuild`, present here. Naming it because the brief requires CI state to be stated precisely.
- **The body's macOS claim: NOT VERIFIED.** All my builds are Linux/GCC.
- **`-DEOS_PRODUCT=medical` and `-DEOS_PRODUCT=aerospace`: NOT BUILT.** I built the default host configuration and `vbox_test` only.
- **`test_crypto_ed25519_loworder`'s correctness: NOT VERIFIED**, only that it builds, runs and passes. I did not check its low-order encodings against the curve, nor that `eos_crypto` rejects them for the right reason.
- **Whether other test files under `tests/` remain unregistered: NOT CHECKED** here. That is #114's subject.
- **The pinned eBoot tree was not built.** I confirmed `a172a6d` exists and is on eBoot's master; I did not compile eos's kernel job against it, so "the pin makes the job green" rests on `checks.txt`, not on anything I ran.

### Evidence
Read-only `git archive` extraction of `2a32e3a`; the `eos` checkout was clean and was not modified.

```
                  -DEOS_PRODUCT=vbox_test && cmake --build          -> OK
            ctest --no-tests=error                                  -> 39/39 passed
            cmake -B build/host -DCMAKE_BUILD_TYPE=Release          -> OK  (default config)

deletion is safe:
            tests/test_net.c @ 2a32e3a       grep -n assert  -> no match
            tests/test_net.c @ origin/master grep -n assert  -> 5:#include <assert.h>   (only)

finding 1:
            git diff 2a32e3a <#135 head> -- .github/workflows/eos-simulation.yml  -> empty
            #135 head is f34edb3, one commit ahead (the staleness guard this copy lacks)

the pin:
            eBoot  a172a6d  "fix(ed25519): reject public keys outside the prime-order
                             subgroup (#86)"  — ancestor of eBoot origin/master
            ebuild e5d8052  "Merge pull request #102"  — ebuild origin/master
```


---
*Automated architecture review of `2a32e3a919c1` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #120, opened 2026-09-01
fix(pkg): give eos_pkg a real trust anchor, or refuse to verify

Kartikey1306:
Closes the second half of #98 — the trust anchor, which @srpatcha called
"still the more important half" after #99 landed the verifier half.

## The state #99 left behind

#99 stopped `services/crypto` accepting low-order keys, which closed the half
where `eos_pkg` accepted every package. The anchor itself is untouched on
`master`:

```c
services/pkg/eos_pkg.c:94
static const uint8_t eos_pkg_public_key[32] = {0};
```

So `eos_pkg` went from accepting everything to rejecting everything. **Verified** —
a package whose payload is RFC 8032 §7.1 Test 2's message, carrying Test 2's real
signature, built against `master` unmodified:

```
== the package is genuinely signed ==
ed25519_verify with the correct key : ACCEPT
ed25519_verify with the zero anchor : reject

== what eos_pkg_verify() reports, with the anchor it ships ==
eos-pkg: signature verification failed
eos_pkg_verify rc = -1
```

The package is fine. The key was never provisioned. The message says the
opposite, so the one diagnostic an operator gets points away from the cause.

## Why nobody could have caught it

**Verified** — `services/pkg/eos_pkg.c` appears in no `CMakeLists.txt`:

```
$ grep -rn "services/pkg\|eos_pkg.c" --include=CMakeLists.txt .
(no matches)
```

Nothing compiled the file. `tests/test_package.c` tests `pkg/src/package.c` —
the build-recipe registry, a different module that happens to own the target
name `eos_pkg`. So the anchor was unreachable by any test, which is the reason
it survived four rounds of review.

It compiles clean as it stands (`-Wall -Wextra -Wpedantic`, 0 warnings); it was
simply never wired in. It is now the `eos_eapp` library.

## The fix

The key comes from the platform, in the shape this repository already uses:

| source | precedent |
|---|---|
| `eos_pkg_set_trust_anchor()` — where a keystore or provisioning step installs it | `eos_ota_set_authenticator()` (#75) |
| `EOS_PKG_TRUST_ANCHOR_HEX` — 64 hex characters from the build | — |
| neither → **refuse**, and say why | `EOS_ALLOW_STUB_CRYPTO` (#84), `EOS_ALLOW_UNSIGNED_OTA` (#75) |

`EOS_ALLOW_UNSIGNED_PKG` is the opt-in, for tests and bring-up.

An anchor is validated **when it is installed**, not once per package:

```
eos-pkg: refusing a trust anchor that is not a prime-order Ed25519 point;
         such a key cannot authenticate anything
```

That needed #99's checks to be reachable from outside `ed25519_verify()`, so
they are exported as `ed25519_public_key_is_usable()` rather than duplicated —
one implementation of the rule, one set of constants.

The three failures are reported apart because they call for different actions:
fix the provisioning, fix the build, or reject the package.

## Tests fail on the unfixed code

`tests/test_pkg_trust_anchor.c`, 8 tests. **Verified** they discriminate — built
against `master`'s `eos_pkg.c` with only the new symbols shimmed in so it links:

```
  test_no_anchor_refuses_a_genuine_package                 [PASS]
  test_low_order_keys_are_refused_as_anchors               [FAIL]
```

and with the anchor-validation half also shimmed in, so the run reaches the
verification path:

```
  test_no_anchor_refuses_a_genuine_package                 [PASS]
  test_low_order_keys_are_refused_as_anchors               [PASS]
  test_a_real_key_is_accepted_as_an_anchor                 [PASS]
  test_a_genuine_package_verifies_under_its_own_key        [FAIL]
```

Both halves of the change are load-bearing and independently proven.

Coverage: all eight low-order encodings refused as anchors — the all-zero value
this file shipped is only one of them; a real key accepted; a genuinely signed
package **accepted** under its own key, so a verifier that rejects everything
does not pass; a different real key rejected; a tampered payload rejected *with
the header hash recomputed over it*, so the SHA-256 check passes and only the
signature can reject it; and a rejected anchor leaving the one in force intact,
so a failed provisioning attempt cannot disarm a working installation.

## Verification

| item | result |
|---|---|
| `cmake -B build/host -DEOS_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release` | PASS |
| `test_pkg_trust_anchor` | PASS 8/8 |
| `eos_pkg.c` under `-Wall -Wextra -Wpedantic` | PASS, 0 warnings |
| ctest, this branch (macOS) | 34/35 |
| ctest, `master` same machine, same flags | 33/34 |

**Verified** — the single failure is `test_net`, identical on both sides and
pre-existing. It is macOS-only: `tests/test_net.c` calls `assert()` without
`#include <assert.h>`, which glibc admits transitively and clang does not. Not
touched here; worth its own one-line fix.

## Left open deliberately

- **Observed, not fixed:** `eos_pkg.c` runs `rm -rf "%s"` / `rmdir /s /q "%s"`
  through `system()` with an interpolated path (`eos_rmdir_recursive`). Now that
  the file is compiled, that is reachable code. It is a separate change and a
  separate argument; flagging it rather than folding it in.
- The `.eapp` payload is signed, the header is not — `entry`-style fields are
  outside the signature, the same shape as eBoot #40. Worth deciding on; not
  part of closing #98.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [Unreleased]
 
 ### Fixed
+- **`eos_pkg` trust anchor:** Package signatures were checked against `eos_pkg_public_key[32] = {0}`. #99 stopped that all-zero key accepting every package, which left it rejecting every package -- including correctly signed ones -- while reporting `signature verification failed`, blaming the package for a key that was never provisioned. The key now comes from `eos_pkg_set_trust_anchor()` or from `EOS_PKG_TRUST_ANCHOR_HEX` at build time; with neither, verification refuses and says so, unless the build defines `EOS_ALLOW_UNSIGNED_PKG`. Closes the second half of #98.
+- **`services/pkg` is compiled.** `services/pkg/eos_pkg.c` was in no `CMakeLists.txt`, so no target built it and no test could reach it -- which is how the all-zero anchor survived. It is now the `eos_eapp` library, with `tests/test_pkg_trust_anchor.c` covering it.
+- **`ed25519_public_key_is_usable()`:** #99's subgroup and identity checks are exported, so a stored trust anchor can be rejected when it is configured rather than once per package as an apparent signature failure.
 - **`eos_queue_send` / `eos_queue_receive`:** A full send/recv waiter table now returns `EOS_KERN_NO_MEMORY` instead of blocking a task that can never be woken, matching mutex and semaphore behavior.
 - **`eos_queue_create`:** Size check now uses division so `item_size * capacity` cannot wrap `size_t` on 32-bit targets and overflow the 1024-byte queue store.
 - **`eos_sem_create`:** Reject `initial > max` and `max` values that do not fit in `int32_t`, so the counting-semaphore invariant cannot be created already broken.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,13 @@ add_subdirectory(systems)
 # Services — crypto, security, os, linux, rtos, sensor, motor, ota, fs
 # ====================================================================
 add_subdirectory(services)
-add_subdirectory(services/pkg)
+# services/pkg is the host-side .eapp package manager: it lists directories,
+# spawns processes and waits on them, so it needs <dirent.h>, <unistd.h> and
+# <sys/wait.h>. Bare-metal newlib has none of those -- <dirent.h> is a hard
+# #error there -- so it is host-only, gated the same way net_posix.c is.
+if(NOT CMAKE_CROSSCOMPILING)
+    add_subdirectory(services/pkg)
+endif()
 add_subdirectory(pkg)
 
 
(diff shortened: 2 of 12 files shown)


Kartikey1306:
Rebased onto #114 (`518eccf`), which carries #113 and the `assert.h` fix.

`master` is currently red for two reasons that have nothing to do with this
change, and the first build of this branch inherited both:

- `Default configuration (no tests, no product)` — `net/src/net_posix.c` has
  zero `EOS_ENABLE_NET` occurrences on `master`, so its body compiles
  unconditionally. Fixed by #113.
- `Build & Test (Linux x86_64)` — `tests/test_net` fails to link; `test_net.c`
  calls `assert()` with no `#include <assert.h>`, which glibc used to admit
  transitively. Fixed by `9d2f7ea`.

Both are in the stack now, so this branch builds on its own.

Local after the rebase: **39/39 ctest**, on macOS, Release.

The stack fast-forwards from `master`
(`git merge-base --is-ancestor origin/master HEAD` is true), so #113 → #114 →
this can merge in order, or the tip alone takes all three.

One thing this rebase made tidier than I expected: #114 is "test suites that
existed but were never compiled" and this is the same defect one level up — a
whole *service* that no `CMakeLists.txt` built. The all-zero trust anchor is
what was hiding inside it.

srpatcha (approved):

PR #130, opened 2026-09-03
Fix task wake timeout at tick zero

Tanisha1723:
Summary

Fixes task wake-up handling when a timeout deadline wraps around the 32-bit tick counter to 0.

Changes
Use wake_armed to determine whether a task has an active timeout instead of checking wake_tick > 0.
Add a regression test covering a timeout whose deadline wraps to tick 0.
Verify that EOS_WAIT_FOREVER tasks remain blocked when wake_tick is 0.
Testing
test_kernel: 13/13 tests passed
git diff --check: clean

### DIFF
--- a/kernel/src/task.c
+++ b/kernel/src/task.c
@@ -174,7 +174,9 @@ void eos_task_wake_check(uint32_t current_tick)
 {
     for (int i = 0; i < EOS_MAX_TASKS; i++) {
         eos_task_t *t = &g_tasks[i];
-        if (t->state == EOS_TASK_BLOCKED && t->wake_tick > 0 &&
+        /* wake_armed, not wake_tick > 0: a deadline of 0 is valid when
+         * g_tick + timeout wraps. WAIT_FOREVER leaves wake_armed clear. */
+        if (t->state == EOS_TASK_BLOCKED && t->wake_armed &&
             tick_expired(current_tick, t->wake_tick)) {
             t->state = EOS_TASK_READY;
             t->wake_armed = 0;
--- a/tests/test_kernel.c
+++ b/tests/test_kernel.c
@@ -260,6 +260,42 @@ static void test_tick_overflow(void) {
     printf("[PASS] tick overflow wraparound\n");
 }
 
+static void test_wake_tick_zero(void) {
+    eos_kernel_init();
+
+    /* g_tick + 1 wraps to 0. wake_armed must still mark that deadline live. */
+    g_tick = 0xFFFFFFFFU;
+    int timed = eos_task_create("wrap0", test_entry, NULL, 5, 1024);
+    assert(timed >= 0);
+    eos_task_block_with_timeout((eos_task_handle_t)timed, 1);
+    assert(eos_task_get_state((eos_task_handle_t)timed) == EOS_TASK_BLOCKED);
+    assert(g_tasks[timed].wake_tick == 0);
+    assert(g_tasks[timed].wake_armed == 1);
+
+    /* One tick before wrap: not expired */
+    eos_task_wake_check(0xFFFFFFFFU);
+    assert(eos_task_get_state((eos_task_handle_t)timed) == EOS_TASK_BLOCKED);
+
+    /* Tick 0 is the deadline — must wake even though wake_tick is 0 */
+    eos_task_wake_check(0);
+    assert(eos_task_get_state((eos_task_handle_t)timed) == EOS_TASK_READY);
+
+    /* WAIT_FOREVER also stores wake_tick = 0, but wake_armed stays 0 */
+    int forever = eos_task_create("forever", test_entry, NULL, 5, 1024);
+    assert(forever >= 0);
+    eos_task_block_with_timeout((eos_task_handle_t)forever, EOS_WAIT_FOREVER);
+    assert(eos_task_get_state((eos_task_handle_t)forever) == EOS_TASK_BLOCKED);
+    assert(g_tasks[forever].wake_tick == 0);
+    assert(g_tasks[forever].wake_armed == 0);
+
+    eos_task_wake_check(0);
+    assert(eos_task_get_state((eos_task_handle_t)forever) == EOS_TASK_BLOCKED);
+    eos_task_wake_check(1);
+    assert(eos_task_get_state((eos_task_handle_t)forever) == EOS_TASK_BLOCKED);
+
+    printf("[PASS] wake_tick wrap to zero\n");
+}
+
 /* Mock port functions for host-based simulation/testing */
 uint32_t eos_port_enter_critical(void) { return 0; }
 void eos_port_exit_critical(uint32_t state) { (void)state; }
@@ -320,6 +356,7 @@ int main(void) {
     test_queue_full();
     test_task_stats();
     test_tick_overflow();
-    printf("=== ALL KERNEL TESTS PASSED (12/12) ===\n");
+    test_wake_tick_zero();
+    printf("=== ALL KERNEL TESTS PASSED (13/13) ===\n");
     return 0;
 }


srpatcha (commented):
## Review — eos#130 "Fix task wake timeout at tick zero"
head: 6e7f4a36c62e0e450196c0d9fa5032e8a38c90b7  author: Tanisha1723  ci: **none — zero checks have run** (fork PR, workflows awaiting maintainer approval; see finding 1)

**Verdict:** Correct, minimal, and the right fix. `wake_armed` already existed in `eos_task_t` and was already maintained at both arming sites; `eos_task_wake_check()` was the one place still inferring "armed" from `wake_tick > 0`, which is also the `EOS_WAIT_FOREVER` sentinel — so a deadline that wrapped exactly to 0 was indistinguishable from "never wake", and the task blocked permanently. One line, using infrastructure already in the tree, plus a regression test that genuinely fails without it. Merge it once CI has actually run.

### Verified by running it
Built the PR head in a scratch worktree and ran the suite:

    cmake -S . -B bt -DEOS_BUILD_TESTS=ON   # rc=0
    cmake --build bt -j4                    # rc=0
    ctest --test-dir bt --no-tests=error    # 38/38 passed
    ctest --test-dir bt -R "test_kernel|test_e2e" --no-tests=error   # both Passed

The author's `test_kernel: 13/13 tests passed` is confirmed.

**And the test is a real regression test.** I reverted only the one-line predicate back to `t->wake_tick > 0`, kept the new test, and rebuilt:

    test_kernel: /tmp/.../tests/test_kernel.c:281: test_wake_tick_zero:
      Assertion `eos_task_get_state((eos_task_handle_t)timed) == EOS_TASK_READY' failed.
    1 test failed out of 1

So it fails against the old code and passes against the new — which is the property that matters and the one PR bodies in this repo usually assert without showing.

I also checked the invariant the change depends on, because switching the gate could have silently *stopped* wakeups rather than started them. It holds: `wake_armed = 1` is written at exactly the two sites that set a live deadline — `eos_task_delay_ms` (`task.c:371-372`) and `eos_task_block_with_timeout` (`task.c:513-514`) — and cleared only where the deadline genuinely dies (`wake_check` on fire, `task.c:182`; the `EOS_WAIT_FOREVER` branch, `task.c:511`; `eos_task_unblock`, `task.c:523`). Slot reuse is safe: `eos_task_create` does `memset(&g_tasks[slot], 0, sizeof(eos_task_t))` (`task.c:303`) and `eos_kernel_init` memsets the whole array (`task.c:218`). `eos_task_delete` leaves the flag set but moves the state to `EOS_TASK_DELETED`, which `wake_check` does not match. There is no path that leaves a task `BLOCKED` with a live deadline and `wake_armed == 0`.

### Findings
|---|---|---|---|---|
| 1 | Medium | — (repo/CI state, not the diff) | **A kernel scheduler change has no CI evidence whatsoever.** `checks.txt` is empty and the API agrees: `commits/6e7f4a36/check-runs` → `total_count: 0`. This is a fork PR (`isCrossRepository: true`, head owner `Tanisha1723`), so the workflow runs are almost certainly sitting behind GitHub's first-time-contributor approval gate, and `mergeStateStatus: BLOCKED` follows from that. Nothing here is the author's fault, but the consequence is that the only evidence for a change to `eos_task_wake_check` is the author's local run and this review's — no cross-compile, no QEMU boot, no static analysis, no CodeQL. | A maintainer clicks **Approve and run workflows** on this PR, and it is not merged until `Build & Test (Linux x86_64)`, `Cross-compile ARM Cortex-M4`, `QEMU Boot Test` and `Static Analysis` have reported. This is the one blocking item. |
| 2 | Low | `kernel/src/task.c:36-39`, `509-521` | **Any timeout above `INT32_MAX` fires immediately, and this PR's test stops just short of it.** `tick_expired(c, d)` is `(int32_t)(c - d) >= 0`, a sliding window that is only correct for deadlines within `2^31` of the current tick. Compiled and measured: with `g_tick = 1000`, timeouts `0x1`, `0x7FFFFFFF` and `0x80000000` are correctly "not expired" at the moment of arming, but `0x80000001` and `0xFFFFFFFE` both report **expired at arming** — so the task wakes on the very next tick instead of ~24.9 days later (at a 1 kHz tick). `EOS_WAIT_FOREVER` (`0xFFFFFFFF`, `kernel.h:49`) is excluded by the explicit branch, so the whole range `[0x80000001, 0xFFFFFFFE]` is silently wrong. Adjacent to this PR's own subject — wrap-boundary tick arithmetic — and untested. | Reject it rather than mis-handle it: in `eos_task_block_with_timeout` and `eos_task_delay_ms`, treat `timeout_ms > INT32_MAX && timeout_ms != EOS_WAIT_FOREVER` as `EOS_KERN_INVALID` (or clamp to `INT32_MAX` and document it), and add the `0x80000001` case to `test_wake_tick_zero` next to the existing wrap case. Document the `2^31` window on `tick_expired`, which currently says only "true if the deadline has passed". |
| 3 | Low | `kernel/include/eos/kernel_internal.h:60-64` vs `kernel/src/task.c:509-521` | `eos_task_block_with_timeout` writes `state`, `wake_tick` and `wake_armed` with no critical section, while its sibling `eos_task_delay_ms` (`task.c:367-372`) takes `eos_port_enter_critical()` for the same three fields — and `eos_task_wake_check` runs from the SysTick ISR (`kernel/src/arch/arm_cm/systick.c:43`). I checked all four production callers and **the race is closed**: `ipc.c:95`, `ipc.c:159`, `sync.c:186` and `sync.c:333` all hold the critical section across the call. So this is not a live defect — it is an undocumented precondition that now carries more weight, because after this PR a torn write to `wake_armed` is the difference between a task waking and hanging. The header comment describes what the function sets and says nothing about who must hold the lock. | One line in `kernel_internal.h:60-64`: state that the caller must hold the critical section, and that `eos_task_delay_ms` is the locking variant. Alternatively make the function self-locking for symmetry — but only if `eos_port_enter_critical` is verified re-entrant on every port, which I did not check. |
| 4 | Low | `tests/test_kernel.c:263-297` | The test covers the wrap-to-zero case through `eos_task_block_with_timeout` only. The other arming site, `eos_task_delay_ms` (`task.c:371`), has the identical exposure and is the one an application calls directly — it is the API behind Appendix A's `threads` and `timer` examples. The fix covers both because both funnel through `wake_check`, so this is a coverage gap, not a correctness gap. | Add four lines to `test_wake_tick_zero`: set `g_tick = 0xFFFFFFFF`, `eos_task_set_current_internal(h)`, `eos_task_delay_ms(1)`, assert `wake_tick == 0 && wake_armed == 1`, then `eos_task_wake_check(0)` and assert `READY`. |
| 5 | Low | `tests/test_kernel.c:359-360` | `printf("=== ALL KERNEL TESTS PASSED (13/13) ===")` hardcodes the count and is bumped by hand with each test. It is printed unconditionally after the last call, so it is only true because every case aborts on `assert`. Pre-existing pattern, and the PR follows it correctly — noted because the number is now the third place the suite size is recorded (here, the PR body, and `tests/CMakeLists.txt`). | Not for this PR. Worth replacing with a counter incremented per case, so the banner cannot claim a count the run did not achieve. |

The diff also drops the missing newline at end of `tests/test_kernel.c`, which is a real if incidental fix and matches the body's `git diff --check: clean` claim.

### Architecture conformance
**Conforms.**

- **§5.1 / §21 Tier 1** — the change is confined to `kernel/src/task.c` and `tests/`. `.ai/architect.md` puts the scheduler, threads and sync primitives in `kernel/`, and this is exactly that. No new include, no link edge, nothing pointing up a tier; `kernel/` gains no dependency on `net/`, `drivers/` or any product repo. §6 is satisfied: "Networking, databases, AI and distributed communication must not be kernel responsibilities" — untouched.
- **§6.1, "Interrupt, timer and synchronization semantics must be documented and testable"** — this is the requirement the PR serves, and the one findings 2 and 3 say is only partly met. The wake path is now testable at the wrap boundary (the new test proves it), but `tick_expired`'s `2^31` window is neither documented nor tested, and `eos_task_block_with_timeout`'s locking precondition is not documented.
- **§6.1, "Deterministic scheduling behavior must be specified by configuration and target"** — a timeout that expires 24.9 days early (finding 2) is a determinism defect in the strict sense, which is why it is worth naming even though it predates this PR.
- **§28 Status and evidence** — the PR body is the good kind for this repo: it states what changed, names the two behaviours asserted, gives a specific test result (`13/13`) and a specific hygiene check (`git diff --check`), and claims nothing further. Both claims verify. The gap is external (finding 1): no CI has run, so §28's `Implemented` evidence bar — "code and functional tests" — is met on the author's host and mine, on x86_64 host builds only.
- **Weakened checks** — none. One test added, no assertion removed, no lint or permission changed, no `|| true`.

An aside, not a finding and not this PR's: `bsan/` and `bsan2/` are tracked at the repository root on `master` and contain compiled test binaries (`bsan/tests/test_kernel` and others match a `git grep` over tracked files). Unrelated to this change; flagging only because committed build output is the kind of thing that stays unnoticed indefinitely.

### Proposed changes
1. **Maintainer: approve the workflow runs** (finding 1). Nothing else should happen first — this is a scheduler change with no cross-compile or QEMU evidence.
2. Merge once those are green. The fix itself needs no change.
3. Same PR if the author wants it, otherwise a follow-up: reject `timeout_ms > INT32_MAX` and add that case to the test (finding 2). This is the one with real behavioural consequence.
4. One-line header comment on the locking precondition (finding 3), and the four-line `eos_task_delay_ms` case in the test (finding 4).

**I opened no fix PR.** Nothing is rated High; finding 2 is the closest to substantive and it is a behavioural change to a kernel timeout API, which `.ai/architect.md` and the brief both put outside what an unattended reviewer should push.

### Not checked
- **Host x86_64 Linux only.** No cross-compile, no `QEMU Boot Test`, no Cortex-M SysTick path. The race analysis in finding 3 is read from `arch/arm_cm/systick.c` and the four call sites; I did not run anything on ARM, and the host test harness supplies mock `eos_port_enter_critical`/`exit_critical` that do nothing (`tests/test_kernel.c:300-301`), so **no test in this PR exercises real critical-section behaviour at all**.
- Because those mocks are no-ops, the concurrency question — SysTick firing inside `eos_task_block_with_timeout` — is not testable by this suite on this host and was not tested. Finding 3's "the race is closed" is from reading `ipc.c` and `sync.c`, not from observation.
- `eos_task_suspend` / `eos_task_resume` interaction with `wake_armed` was not traced. If `resume` restores a task to `EOS_TASK_BLOCKED` the armed deadline survives, and if it restores `READY` the timeout is silently dropped; either way it is pre-existing and I did not read those functions.
- Only `test_kernel`, `test_e2e` and the full 38-test `ctest` run were exercised. `tests/main_system_test.c:111` also reads `wake_tick` directly; it built and the suite is green, but I did not read it.
- No timing or interrupt-latency measurement — §6.1's determinism claims are unmeasured here, and no benchmark exists in this PR to re-run.
- I did not check whether `EOS_MAX_TASKS` scanning in `wake_check` is bounded appropriately for the SysTick budget; the loop is unchanged by this PR.
- The reason CI has not run is inferred from `isCrossRepository: true` plus `total_count: 0` — consistent with GitHub's fork-approval gate, but I could not read the repository's Actions settings to confirm that is the specific gate.


---
*Automated architecture review of `6e7f4a36c62e` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (dismissed):

PR #131, opened 2026-09-03
fix(kernel): preserve scheduler task invariants

piyushkr0509:
## Summary

This PR fixes task selection and stack-pointer bookkeeping in the Cortex-M scheduler, and protects the kernel-owned idle task. I chose these two areas for the technical assessment after reviewing task creation, scheduling, and the PendSV save/restore sequence.

On the base commit (`feee272`), unused task slots are zero-initialized, which makes them appear `READY` at priority 0. The scheduler can therefore select an empty slot over a real task and hand the restore path a null stack pointer. Separately, `g_current_sp` keeps pointing at the outgoing task after a switch. On the following PendSV, the running task's stack pointer would be saved into the previous task's control block.

The scheduler now selects initialized, runnable tasks and keeps the stack-pointer slot aligned with the task actually running. It also keeps a preempted task eligible at its own priority and allows priority 255, so idle can run when application tasks are blocked.

## Before and after

The [review of this commit](https://github.com/embeddedos-org/eos/pull/131#pullrequestreview-5104656159) independently reproduced the selection failures using host probes:

| Scenario | Base (`feee272`) | This PR (`e291bee`) |
| --- | --- | --- |
| Initialize the kernel, create one task at priority 5, then call `eos_schedule()` | Selects empty slot 2: `name=invalid`, `stack_ptr=(nil)` | Selects the real task in slot 1 with a non-null saved stack pointer |
| Block all application tasks | Selects an empty slot instead of idle; the strict priority comparison also excludes priority 255 | Selects slot 0: `name=idle` |

For the stack-pointer handoff, tracing A → B → A shows that the old code leaves `g_current_sp` pointing at A after selecting B. This PR updates it to B before returning, so the next save goes to B's own TCB. The host regression simulates that save and checks that A's saved pointer is unchanged. Execution of the ARM handler itself remains unverified, as noted below.

## Type of change

- [x] fix — Bug fix
- [x] test — Add regression tests

## Implementation

- Reject deletion and suspension of the idle task in reserved slot 0 with `EOS_KERN_INVALID`.
- Exclude slots with a null entry function and accept priority 255 during task selection.
- Check the outgoing task's canary before selection, and keep a still-running task eligible alongside other runnable tasks. Equal-priority round-robin ordering is preserved.
- Point both `g_current_sp` and `g_next_sp` at the selected task's saved stack-pointer slot, and document the C/assembly contract in `kernel_internal.h`.
- Protect slot selection, static stack allocation, and TCB initialization with a critical section. Publish the entry function and `READY` state after initialization, restore the previous interrupt state on every exit, and request preemption after leaving the critical section.
- Use `push {r3, lr}` / `pop {r3, lr}` around the call to `eos_schedule()` to preserve 8-byte stack alignment at the C call boundary.
- Add regression coverage for idle protection, scheduling during task initialization, priority retention, consecutive stack-pointer handoffs, and idle fallback.

Public function signatures and task structure layout are unchanged. The intentional API behavior change is that deleting or suspending the idle task is now rejected.

## Testing

Results for commit `e291bee` on the macOS ARM64 host:

- Kernel test executable: **14/14 passed**.
- Existing registered host suite, Debug build: **38/38 passed**.
- Existing registered host suite with AddressSanitizer and UndefinedBehaviorSanitizer: **38/38 passed** (`detect_leaks=0`).
- Release library build: **passed**.
- Strict C11 compilation of `task.c`: **passed**.
- Cortex-M4 assembly compilation of `context_switch.S`: **passed**.

The reviewer independently reproduced the 14/14 kernel and 38/38 host/sanitizer results. The new tests exercise scheduler behavior through the host port mocks; they do not execute real interrupt masking or ARM exception entry/return.

## Validation still needed

At the time of this description update, GitHub reports no checks for this PR. Maintainer approval of the fork workflow runs is needed before relying on CI results.

The current QEMU boot job omits `context_switch.S` and `systick.c`, and its system test does not call `eos_kernel_start()`. Neither the host tests nor the assembly compilation demonstrate a running PendSV context switch. A QEMU or hardware test that starts the scheduler and performs repeated handoffs is still needed to validate that path end to end.

## Related issues

No issue is linked; these changes came from the assessment code review. The commit follows the conventional commit format and includes a DCO sign-off.

### DIFF
--- a/kernel/include/eos/kernel_internal.h
+++ b/kernel/include/eos/kernel_internal.h
@@ -23,6 +23,19 @@ extern volatile uint32_t g_tick;
 /* Task array — accessible to sync.c and ipc.c for priority inheritance */
 extern eos_task_t g_tasks[EOS_MAX_TASKS];
 
+/* Stack-pointer slots shared with the architecture context-switch handler. */
+extern uint32_t **g_current_sp;
+extern uint32_t **g_next_sp;
+
+/**
+ * @brief Select the next runnable task during a context switch.
+ *
+ * The port saves the outgoing CPU stack pointer through g_current_sp before
+ * calling this function. On return, both stack-pointer slots identify the
+ * newly running task so the port can restore it and save it on the next switch.
+ */
+void eos_schedule(void);
+
 /**
  * @brief True once @p now has reached or passed @p deadline.
  *
--- a/kernel/src/arch/arm_cm/context_switch.S
+++ b/kernel/src/arch/arm_cm/context_switch.S
@@ -45,9 +45,9 @@ PendSV_Handler:
     str     r0, [r1]            /* Save PSP */
 
     /* ---- Select next task ---- */
-    push    {lr}
+    push    {r3, lr}            /* Preserve 8-byte AAPCS stack alignment */
     bl      eos_schedule        /* Updates g_current_sp and g_next_sp */
-    pop     {lr}
+    pop     {r3, lr}
 
     /* ---- Restore next context ---- */
     ldr     r1, =g_next_sp
--- a/kernel/src/task.c
+++ b/kernel/src/task.c
@@ -125,8 +125,8 @@ static int find_next_task(void)
     for (int i = 0; i < EOS_MAX_TASKS; i++) {
         int idx = (start + i) % EOS_MAX_TASKS;
         eos_task_t *t = &g_tasks[idx];
-        if (t->state != EOS_TASK_READY) continue;
-        if (t->priority < best_prio) {
+        if (t->entry == NULL || t->state != EOS_TASK_READY) continue;
+        if (best < 0 || t->priority < best_prio) {
             best_prio = t->priority;
             best = idx;
         }
@@ -141,15 +141,7 @@ static int find_next_task(void)
  */
 void eos_schedule(void)
 {
-    int next = find_next_task();
-    if (next < 0) return;  /* No ready tasks — stay on current */
-
-    /* Mark current task as ready (if it was running) */
-    if (g_current >= 0 && g_tasks[g_current].state == EOS_TASK_RUNNING) {
-        g_tasks[g_current].state = EOS_TASK_READY;
-    }
-
-    /* Check stack canary on current task */
+    /* Validate the outgoing task before making it eligible for selection. */
     if (g_current >= 0 && g_tasks[g_current].stack_base) {
         if (g_tasks[g_current].stack_base[0] != EOS_STACK_CANARY) {
             /* Stack overflow detected! */
@@ -158,12 +150,24 @@ void eos_schedule(void)
         }
     }
 
-    /* Switch to next task */
-    g_current_sp = &g_tasks[g_current >= 0 ? g_current : next].stack_ptr;
+    /* A preempted task remains runnable and competes at its normal priority. */
+    if (g_current >= 0 && g_tasks[g_current].state == EOS_TASK_RUNNING) {
+        g_tasks[g_current].state = EOS_TASK_READY;
+    }
+
+    int next = find_next_task();
+    if (next < 0) return;  /* No valid task is runnable. */
+
+    /*
+     * PendSV has already saved the outgoing PSP through g_current_sp.
+     * Point both assembly-facing slots at the task that is now current, so
+     * the next PendSV saves that task's PSP into its own control block.
+     */
     g_current = next;
     g_tasks[next].state = EOS_TASK_RUNNING;
     g_tasks[next].run_count++;
-    g_next_sp = &g_tasks[next].stack_ptr;
+    g_current_sp = &g_tasks[next].stack_ptr;
+    g_next_sp = g_current_sp;
 }
 
 /**
@@ -281,6 +285,8 @@ int eos_task_create(const char *name, eos_task_func_t entry, void *arg,
     if (!entry) return EOS_KERN_INVALID;
     if (stack_size == 0) stack_size = 1024;
 
+    uint32_t crit = eos_port_enter_critical();
+
     /* Find free slot (skip slot 0 = idle) */
     int slot = -1;
     for (int i = 1; i < EOS_MAX_TASKS; i++) {
@@ -290,30 +296,41 @@ int eos_task_create(const char *name, eos_task_func_t entry, void *arg,
             break;
         }
     }
-    if (slot < 0) return EOS_KERN_NO_MEMORY;
+    if (slot < 0) {
+        eos_port_exit_critical(crit);
+        return EOS_KERN_NO_MEMORY;
+    }
 
     /* Allocate stack */
     uint32_t *stack_base = alloc_task_stack(stack_size);
-    if (!stack_base) return EOS_KERN_NO_MEMORY;
+    if (!stack_base) {
+        eos_port_exit_critical(crit);
+        return EOS_KERN_NO_MEMORY;
+    }
 
     uint32_t *stack_top = stack_base + stack_size / sizeof(uint32_t);
 
     memset(&g_tasks[slot], 0, sizeof(eos_task_t));
+    g_tasks[slot].state = EOS_TASK_DELETED;
     g_tasks[slot].id = (uint8_t)slot;
     g_tasks[slot].name = name;
-    g_tasks[slot].entry = entry;
     g_tasks[slot].arg = arg;
     g_tasks[slot].priority = priority;
     g_tasks[slot].base_priority = priority;
     g_tasks[slot].stack_size = stack_size;
     g_tasks[slot].stack_base = stack_base;
     g_tasks[slot].stack_ptr = eos_port_init_stack(stack_top, entry, arg);
+
+    /* Publish the entry and READY state only after the TCB is complete. */
+    g_tasks[slot].entry = entry;
     g_tasks[slot].state = EOS_TASK_READY;
     g_task_count++;
 
     /* If kernel is running and new task has higher priority, yield */
-    if (g_running && g_current >= 0 &&
-        priority < g_tasks[g_current].priority) {
+    bool should_preempt = g_running && g_current >= 0 &&
+                          priority < g_tasks[g_current].priority;
+    eos_port_exit_critical(crit);
+    if (should_preempt) {
         eos_port_yield();
     }
 
@@ -322,7 +339,7 @@ int eos_task_create(const char *name, eos_task_func_t entry, void *arg,
 
 int eos_task_delete(eos_task_handle_t h)
 {
-    if (h >= EOS_MAX_TASKS) return EOS_KERN_INVALID;
+    if (h == 0 || h >= EOS_MAX_TASKS) return EOS_KERN_INVALID;
     if (g_tasks[h].state == EOS_TASK_DELETED || g_tasks[h].entry == NULL)
         return EOS_KERN_INVALID;
     g_tasks[h].state = EOS_TASK_DELETED;
@@ -337,7 +354,8 @@ int eos_task_delete(eos_task_handle_t h)
 
 int eos_task_suspend(eos_task_handle_t h)
 {
-    if (h >= EOS_MAX_TASKS || g_tasks[h].entry == NULL) return EOS_KERN_INVALID;
+    if (h == 0 || h >= EOS_MAX_TASKS || g_tasks[h].entry == NULL)
+        return EOS_KERN_INVALID;
     g_tasks[h].state = EOS_TASK_SUSPENDED;
     if (h == (eos_task_handle_t)g_current) {
         eos_port_yield();
@@ -572,5 +590,3 @@ int eos_task_get_all_stats(eos_task_stats_t *out, int max_entries, int *count)
     *count = n;
     return EOS_KERN_OK;
 }
-
-
--- a/tests/test_kernel.c
+++ b/tests/test_kernel.c
@@ -9,7 +9,14 @@
 #include "eos/kernel_internal.h"
 
 static int g_task_ran = 0;
+static uint32_t g_critical_depth = 0;
+static void (*g_stack_init_hook)(void) = NULL;
 static void test_entry(void *arg) { g_task_ran = 1; (void)arg; }
+
+static void schedule_during_stack_init(void) {
+    assert(g_critical_depth > 0);
+    eos_schedule();
+}
 
 static void test_kernel_init(void) {
     assert(eos_kernel_init() == EOS_KERN_OK);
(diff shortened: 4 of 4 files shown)


srpatcha (commented):
## Review — eos#131 "fix(kernel): preserve scheduler task invariants"
head: e291bee386a51f7ea535342cc39f6d25b99768c3  author: piyushkr0509  ci: **none — zero checks have run** (fork PR; see finding 1)

**Verdict:** The changes are correct, and the PR body badly undersells them. This is not invariant hygiene — two of these fix defects that make the Cortex-M preemptive scheduler non-functional on `master`, and I reproduced both. Everything I could execute on a host passes, including under ASan/UBSan. The two problems are that the highest-risk change (the ARM context-switch pairing) has no execution path anywhere in this repo's CI, and that nine behavioural changes are landing as one 148-line commit touching assembly.

### The two defects on master, reproduced

**1. `find_next_task()` prefers an empty slot to every real task.** `EOS_TASK_READY = 0` (`kernel/include/eos/kernel.h:69`) and `eos_kernel_init` memsets all 16 slots, so each of the 15 unused ones reads as `READY` with `priority = 0` — the highest priority in the system. Master's predicate `if (t->priority < best_prio)` therefore selects one. Probe compiled and run against `origin/master` (`feee272`), one real task created at priority 5:

    PROBE: slot2 state=0 prio=0 entry=(nil) sp=(nil)
    PROBE: after eos_schedule g_current=2 name=invalid stack_ptr=(nil)
    PROBE: g_next_sp=0x... *g_next_sp=(nil)

The scheduler chose empty slot 2 over the real task, and `g_next_sp` dereferences to `NULL` — which is the value PendSV loads into PSP. Same probe on this PR's head:

    PROBE: after eos_schedule g_current=1 name=real stack_ptr=0x...
    PROBE: g_next_sp=0x... *g_next_sp=0x...

`eos_kernel_start` (`task.c:255`) also calls `find_next_task()`, so on master this fires at the very first switch, before any user code runs.

**2. Priority 255 could never be selected, so idle was not a fallback.** `best_prio` starts at `255` and the test was strict `<`. With every real task blocked, master returns `-1` and `eos_schedule` leaves the blocked task current. Same probe, one task blocked `EOS_WAIT_FOREVER`:

    master:  PROBE: all-real-tasks-blocked -> g_current=2   (empty slot again)
    this PR: PROBE: all-real-blocked      -> g_current=0 name=idle

**3. `g_current_sp` was paired one switch late.** Read against the handler: `context_switch.S:43-45` saves the outgoing PSP *through* `g_current_sp` **before** `bl eos_schedule`, so `g_current_sp` must already point at the running task on entry. Master's `eos_schedule` sets `g_current_sp = &g_tasks[g_current].stack_ptr` while `g_current` is still the *outgoing* task, then advances `g_current`. So after switch N it points at the task switched **out**; at switch N+1 PendSV saves the running task's PSP into that task's TCB instead of its own, and the running task's `stack_ptr` keeps a stale value. From PendSV #2 onward every task's stack pointer is written into the previous task's control block. This PR's `g_current_sp = &g_tasks[next].stack_ptr; g_next_sp = g_current_sp` makes the save at N+1 land in the right TCB. Traced through `eos_kernel_start` → PendSV #1 → PendSV #2; the invariant holds under the PR and does not under master.

**4. AAPCS alignment.** `push {lr}` left MSP at 4 mod 8 across a public-interface `bl`; `push {r3, lr}` restores 8-byte alignment. Standard idiom, and `r3` is caller-saved and already in the exception frame, so clobbering and restoring it is harmless.

### What I ran

    cmake -S . -B bt -DEOS_BUILD_TESTS=ON && cmake --build bt -j4   # rc=0
    ctest --test-dir bt --no-tests=error                            # 38/38 passed
    ./bt/tests/test_kernel                                          # === ALL KERNEL TESTS PASSED (14/14) ===

    cmake -S . -B bsanx -DEOS_BUILD_TESTS=ON \
      -DCMAKE_C_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer -g" \
      -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined" && cmake --build bsanx -j4
    ASAN_OPTIONS=detect_leaks=0 UBSAN_OPTIONS=print_stacktrace=1 \
      ctest --test-dir bsanx --no-tests=error                       # 38/38 passed

The body's `14/14`, `38/38` and ASan/UBSan `38/38` all check out.

### Findings
|---|---|---|---|---|
| 1 | High | — (CI state) | **Zero checks have run on a scheduler and context-switch change.** `checks.txt` is empty and `commits/e291bee3/check-runs` → `total_count: 0`. Fork PR (`isCrossRepository: true`, head owner `piyushkr0509`), so the runs are behind GitHub's contributor-approval gate. No cross-compile, no QEMU, no CodeQL, no static analysis has seen this. | A maintainer approves the workflow runs, and this does not merge until at least `Cross-compile ARM Cortex-M4`, `QEMU Boot Test`, `Static Analysis` and `Build & Test (Linux x86_64)` have reported. |
| 2 | High | `.github/workflows/build.yml:98` vs `:150-156`; `kernel/src/arch/arm_cm/context_switch.S:48-50` | **`PendSV_Handler` is never executed by any CI job, on master or here — so the riskiest change in this PR cannot be validated.** `Cross-compile ARM Cortex-M4` assembles and links `context_switch.S` (`build.yml:98`) but only runs `objcopy` and `size`. `QEMU Boot Test` is the one job that *runs* an image, and its object list (`build.yml:150-156`) omits `context_switch.S` entirely — it compiles `task.c`, `sync.c`, `ipc.c`, `heap.c`, `nvic.c`, `port.c` and `tests/main_system_test.c`. And no test in the tree calls `eos_kernel_start()` (`grep -rl eos_kernel_start tests/` → nothing), so the switch is never entered even where it is linked. This is also why both defects above have sat on green CI. The author's evidence, `clang --target=arm-none-eabi -c context_switch.S`, proves it assembles — which was not in question. | Add `kernel/src/arch/arm_cm/context_switch.S` and `arch/arm_cm/systick.c` to the QEMU image's object list, and give `main_system_test.c` (or a new `tests/qemu_switch_test.c`) two tasks plus an `eos_kernel_start()` so PendSV actually fires. Assert that each task's `stack_ptr` after a handoff is within its own stack region — that single assertion catches both the pairing bug and the alignment bug, and would have caught them on master. |
| 3 | Medium | whole PR (4 files, +148/−30) | **Nine independent behavioural changes in one commit**, spanning scheduler selection, TCB publication order, task lifecycle validation, critical-section scope and ARM assembly. `.ai/architect.md` is explicit: do not "restructure and change behaviour in the same commit", and state migration impact and the "smallest sequence of steps that keeps every target building throughout". A future bisect onto a scheduler regression lands on 148 lines including hand-written assembly, and a reviewer has to hold all nine in their head at once. | Split, and land in this order: **(a)** `find_next_task` validity + priority-255 — the two reproduced Critical fixes, smallest and most urgent; **(b)** the `g_current_sp` pairing fix plus the `kernel_internal.h` declarations; **(c)** the AAPCS `push {r3, lr}`; **(d)** idle-task protection in `delete`/`suspend`; **(e)** the create-path critical section and publication order. Each is independently testable and (a) and (b) should not wait on the rest. |
| 4 | Medium | PR body, "Summary" and "Changes" | **The body describes the fixes at the wrong severity.** "Preserving scheduler and task-control-block invariants" and "Exclude unused and partially initialized task slots from scheduler selection" read as tidy-up; the evidence above is a NULL PSP load at the first switch and systematic stack-pointer misfiling thereafter. There is also no linked issue and no reproduction, so a reviewer has nothing to check the claims against — which per the brief is itself the finding: nine assertions, none with a failing-before/passing-after demonstration. | Lead the body with the two reproduced defects, each with the before/after evidence (the probe above is reusable), and say plainly that the Cortex-M preemptive path does not work on master. Note explicitly that the assembly change is unverified by execution (finding 2) rather than leaving `clang -c` to imply otherwise. |
| 5 | Low | `kernel/src/task.c:131` | The new `t->entry == NULL` guard is the load-bearing half of fix 1 and its reason is invisible in the code. `state != EOS_TASK_READY` looks like it already covers deleted slots (it does — `eos_task_delete` sets both), so the `entry` check reads as redundant belt-and-braces and is exactly the kind of line a later cleanup deletes. Its actual job is the *never-initialised* slot, where `state == EOS_TASK_READY` because that enumerator is `0`. | One comment on that line: `/* A zeroed slot reads as READY (EOS_TASK_READY == 0) at priority 0, the highest — entry is the only field that distinguishes "never created". */` Cheap, and it is the difference between this fix surviving and being reverted by accident. |
| 6 | Low | `kernel/src/task.c:169-170` | After this change `g_current_sp` and `g_next_sp` are always equal — both on entry to PendSV and on return from `eos_schedule`. Two symbols naming one value is what allowed the original off-by-one to look plausible, and `g_next_sp` no longer means "next". The restore path at `context_switch.S:53-55` could read `g_current_sp`. | Not in this PR. Follow-up: collapse to a single `g_active_sp`, or document at `kernel_internal.h:27-28` why two names are retained. Do it after finding 2's test exists, not before. |
| 7 | Low | `kernel/src/task.c:288-330` | The new critical section spans `alloc_task_stack()` and `eos_port_init_stack()`, so the longest interrupts-disabled window in task creation now includes a stack allocation and a full frame initialisation. Correct, but §6.1 makes deterministic behaviour a specifiable property, and only the *slot claim* needs to be atomic. | Optional, and it needs care: claiming the slot then releasing the lock requires a sentinel the free-slot scan rejects — `entry == NULL || state == EOS_TASK_DELETED` currently matches a half-claimed slot, so a concurrent `eos_task_create` would claim it too. A dedicated `EOS_TASK_CREATING` state would do it. If that is not wanted, leave as is and record the interrupt-latency cost. |
| 8 | Low | `kernel/src/task.c:144-150` | The stack-overflow branch this PR moves still discards its own result: it sets `state = EOS_TASK_DELETED` and returns, with `/* In production: trigger fault handler or reboot */` standing in for the action. A detected stack overflow is therefore silently swallowed — no log, no fault, no notification to update logic — which is `.ai/security.md`'s "a security step whose result is discarded is a finding". Unlike `eos_task_delete` it also leaves `entry` set and does not decrement `g_task_count`; that half is currently inconsequential, since `g_task_count` is `static` and no code reads it for a decision, but it is still wrong. | Pre-existing and out of scope for the split above, but this PR is the one touching these lines. Either call a weak `eos_stack_overflow_hook(handle)` that defaults to the port's fault path, or at minimum log through the core logging service, and mirror `eos_task_delete`'s bookkeeping. |

The new tests are good. `test_scheduler_selects_valid_task_and_tracks_stack` asserts the `g_current_sp`/`g_next_sp` identity after each switch, checks that a tick does not displace the highest-priority runnable task, and covers the priority-255 idle fallback. Making the mock `eos_port_enter_critical`/`exit_critical` count nesting depth (`tests/test_kernel.c:338-346`) and driving `eos_schedule()` from inside `eos_port_init_stack` via `g_stack_init_hook` is a genuinely clever way to prove the publication ordering — worth noting that on real hardware the critical section makes that interleaving impossible, so the test is exercising the belt while the braces are what actually hold.

### Architecture conformance
**Conforms.**

- **§5.1 / §21 Tier 1** — everything is in `kernel/` and `tests/`, which `.ai/architect.md` names as the home of the scheduler, threads and sync primitives, plus `kernel/src/arch/arm_cm/` for the port. §6's boundary is respected: no networking, database, AI or distributed communication enters the kernel. Nothing points up a tier; no product repo is referenced.
- **§6.1, "Architecture ports should expose a stable internal porting contract"** — this is the clause the PR most directly serves. `g_current_sp`, `g_next_sp` and `eos_schedule()` were the contract between `task.c` and `context_switch.S` and were declared in neither header; the port reached them by implicit external linkage. Adding them to `kernel_internal.h:26-37` with the save/restore protocol written down is the right change, and the missing declaration is plausibly how the pairing drifted in the first place. Finding 6 is that the contract is now one symbol wider than it needs to be.
- **§6.1, "Interrupt, timer and synchronization semantics must be documented and testable"** — advanced by the new tests, and finding 2 is the part still unmet: the switch itself remains untestable in this repo's CI.
- **§6.1, "Deterministic scheduling behavior must be specified by configuration and target"** — the priority-255 fix and the "preempted task competes at its own priority" reorder both move toward a specifiable policy. Note the round-robin property survives: `find_next_task` scans from `(g_current + 1) % EOS_MAX_TASKS`, so the re-READY incumbent is visited last and strict `<` lets an equal-priority peer keep `best`. Finding 7 is the determinism cost on the create path.
- **§22 Hardware support tiers** — relevant to how this is described. `Reference` tier requires "hardware CI, release validation". Given finding 2, no Cortex-M target can currently support a `Reference` claim for preemptive scheduling, before or after this PR.
- **Weakened checks** — none. Two tests added, none disabled, no assertion removed, no lint or permission changed, no `|| true`. The mock port functions became *stricter* (they now assert on unbalanced critical sections).

No master-design proposal from this PR: §6.1 already requires everything violated here, so the design was right and the code was wrong.

### Proposed changes
1. **Maintainer: approve the workflow runs** (finding 1). Nothing else first.
2. **Add the QEMU switch test** (finding 2). This is the one thing that would turn "reasoned correct" into "demonstrated correct" for the assembly change, and it is a change to `build.yml` plus a two-task harness — modest, and it protects every future edit to this file.
3. **Split the PR** (finding 3), landing the `find_next_task` fix first. If splitting is refused, land it whole after step 2 — the fixes are right and master is worse — but say in the body that steps 1 and 2 were the gate.
4. Rewrite the body at the correct severity, with the reproduction (finding 4).
5. Add the one-line comment at `task.c:131` (finding 5). Do this regardless of anything else.
6. Findings 6, 7 and 8 as follow-ups, in that order of cheapness.

**I opened no fix PR.** Findings 1 and 2 are both High, and neither is the kind the brief permits an unattended agent to push: finding 1 needs a human to click a button in the GitHub UI, and finding 2 is a new CI job plus a new on-target test — squarely "never a fix you cannot verify by running something", since I have no ARM host and no way to run the QEMU job here.

### Not checked
- **Nothing was executed on ARM, in QEMU, or on hardware.** The context-switch analysis — the central claim of this review — is read from `context_switch.S:30-66` and traced against `eos_schedule` and `eos_kernel_start`. It is **Inferred, not Verified**. I did not assemble the file, did not link an image, and did not run PendSV once.
- The AAPCS alignment fix is likewise unexecuted. I did not check MSP alignment on exception entry for this port's startup code (`boards/startup_stm32f407.S`), so "master left MSP at 4 mod 8 across the `bl`" assumes MSP was 8-aligned beforehand; I did not verify that.
- **The host tests cannot exercise critical sections.** `tests/test_kernel.c:338-346` replaces `eos_port_enter_critical`/`exit_critical` with a depth counter — real interrupt masking is never engaged — so finding 7 and the create-path concurrency claims are reasoned, not observed.
- The two probe programs I ran are scratch additions to `tests/test_kernel.c` in throwaway worktrees at `origin/master` and this head. They are not in the repo and nothing was pushed. Both worktrees have been removed.
- FPU context handling (`vstmdbeq`/`vldmiaeq`, `tst lr, #0x10`) is untouched by this PR and unexamined. The Cortex-M4 cross build uses `-mfloat-abi=hard`; whether the lazy-stacking path interacts with the alignment change was not analysed.
- `eos_task_resume` was not read, so whether it interacts correctly with the new `EOS_TASK_SUSPENDED` handling for the idle task is unknown. The new `test_idle_task_is_permanent` covers `delete` and `suspend` on slot 0 but not `resume`.
- SMP: `kernel/src/multicore.c` compiles into the same library and was not examined. Every claim here is single-core.
- No timing or interrupt-latency measurement, so §6.1's determinism properties are unquantified and finding 7's cost is unmeasured — there is no benchmark in this repo to re-run for it.
- I did not determine why `QEMU Boot Test` is green on master beyond establishing that it omits `context_switch.S` and never calls `eos_kernel_start()`; `tests/main_system_test.c` was read only for those two facts.


---
*Automated architecture review of `e291bee386a5` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (approved):

PR #138, opened 2026-09-04
Fix: eos_net/test_net fail to build on native Windows

ak-sharvesh:
## Summary

Fixes eos_net and test_net failing to build on native Windows. Both included POSIX-only headers unconditionally whenever CMAKE_CROSSCOMPILING was false, which is also true for a native Windows build.

## Type of Change

- [x] fix — Bug fix

## Changes

- Added `AND NOT WIN32` to the `net_posix.c` source gate in root `CMakeLists.txt`
- Narrowed the same `AND NOT WIN32` gate around only the POSIX-specific inner block of `test_net` in `tests/CMakeLists.txt` (compile definitions, pthreads link, `EOS_NET_POSIX`), keeping `add_executable`/`target_link_libraries(eos_net)`/`add_test` unconditional so the 16 non-POSIX test cases still build and run on Windows. (Revised per review — the original version wrongly gated the whole target and dropped those 16 cases.)

## Testing

- [x] Unit tests pass (ctest --test-dir build --output-on-failure)
- [ ] Integration tests pass

Verification output (Windows, MSVC / Visual Studio 2026 Build Tools):

Test #30: test_net ....................... Passed 0.03 sec
84% tests passed, 6 tests failed out of 38

The 6 failures are believed pre-existing and unrelated to this change (not verified against a baseline, since the pre-patch tree doesn't configure at all on Windows). Causes are independent of this diff — missing unistd.h, GCC-only __attribute__ syntax, and unrelated linker symbols — and none of these targets (test_os_adapter/test_config/test_hal/test_e2e/test_pkg_trust_anchor/test_stress) link eos_net.

### Details

`net/src/net_posix.c` and `tests/test_net.c` both include POSIX-only headers (`<netdb.h>`, `<sys/socket.h>`, `<arpa/inet.h>`). They were gated only on `NOT CMAKE_CROSSCOMPILING`, which assumes "not cross-compiling" implies a POSIX host. That's true for Linux/macOS but not for a native Windows build, where `CMAKE_CROSSCOMPILING` is also `FALSE`.

This caused a hard compile failure (`C1083: Cannot open include file: 'netdb.h'`) on Windows, and originally `test_net` disappeared from `ctest` output entirely as "Not Run."

The root `CMakeLists.txt` change narrows an existing gate and has no impact elsewhere. The `tests/CMakeLists.txt` change adds a gate around only the POSIX-specific inner block (not the whole target), so `test_net` now builds and runs everywhere, with only its POSIX round-trip cases skipped on Windows.

Configure and build succeeded on Windows; 6 pre-existing, unrelated targets fail to compile/link on MSVC (see above), and `test_net` builds and passes all 19 cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,7 @@ add_library(eos_net STATIC
 # it, so that condition is true even for a Cortex-M build. hal_linux.c happens
 # to survive that because it uses only portable headers; net_posix.c needs
 # <netdb.h> and <sys/socket.h>, which bare-metal newlib does not have.
-if(NOT CMAKE_CROSSCOMPILING)
+if(NOT CMAKE_CROSSCOMPILING AND NOT WIN32)
     target_sources(eos_net PRIVATE net/src/net_posix.c)
     target_compile_definitions(eos_net PRIVATE EOS_NET_POSIX=1)
 endif()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -190,16 +190,17 @@ target_link_libraries(test_power PRIVATE eos_power)
 add_test(NAME test_power COMMAND test_power)
 
 # --- test_net: Networking abstraction ---
-add_executable(test_net test_net.c)
-target_link_libraries(test_net PRIVATE eos_net)
-# The POSIX round-trip cases need the same switch the library was built with,
-# plus pthreads for the listener they connect to.
-if(NOT CMAKE_CROSSCOMPILING)
+# Only built on genuine POSIX hosts. test_net.c includes <arpa/inet.h>
+# directly (POSIX-only), matching net_posix.c's own WIN32 gate in the
+# top-level CMakeLists.txt.
+if(NOT CMAKE_CROSSCOMPILING AND NOT WIN32)
+    add_executable(test_net test_net.c)
+    target_link_libraries(test_net PRIVATE eos_net)
     target_compile_definitions(test_net PRIVATE EOS_NET_POSIX=1)
     find_package(Threads REQUIRED)
     target_link_libraries(test_net PRIVATE Threads::Threads)
+    add_test(NAME test_net COMMAND test_net)
 endif()
-add_test(NAME test_net COMMAND test_net)
 
 # --- test_motor_ctrl: Motor control PID ---
 add_executable(test_motor_ctrl test_motor_ctrl.c)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -190,17 +190,18 @@ target_link_libraries(test_power PRIVATE eos_power)
 add_test(NAME test_power COMMAND test_power)
 
 # --- test_net: Networking abstraction ---
-# Only built on genuine POSIX hosts. test_net.c includes <arpa/inet.h>
-# directly (POSIX-only), matching net_posix.c's own WIN32 gate in the
-# top-level CMakeLists.txt.
+add_executable(test_net test_net.c)
+target_link_libraries(test_net PRIVATE eos_net)
+# The POSIX round-trip cases need the same switch the library was built with,
+# plus pthreads for the listener they connect to. NOT WIN32 for the same
+# reason as the eos_net gate in the top-level CMakeLists.txt: a native
+# Windows build is not cross-compiling but has no <arpa/inet.h> either.
 if(NOT CMAKE_CROSSCOMPILING AND NOT WIN32)
-    add_executable(test_net test_net.c)
-    target_link_libraries(test_net PRIVATE eos_net)
     target_compile_definitions(test_net PRIVATE EOS_NET_POSIX=1)
     find_package(Threads REQUIRED)
     target_link_libraries(test_net PRIVATE Threads::Threads)
-    add_test(NAME test_net COMMAND test_net)
 endif()
+add_test(NAME test_net COMMAND test_net)
 
 # --- test_motor_ctrl: Motor control PID ---
 add_executable(test_motor_ctrl test_motor_ctrl.c)


srpatcha (commented):
## Review — eos#138 "Fix: eos_net/test_net fail to build on native Windows"
head: 5ed015763799028351bc0cca8c19b9d0d8198311  author: ak-sharvesh  ci: pending (no check runs reported in the bundle; `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED`)

**Verdict:** The root-`CMakeLists.txt` half is correct and fixes a real MSVC build break. The `tests/CMakeLists.txt` half over-corrects: it deletes the whole `test_net` target on Windows when only the POSIX sub-suite needs excluding, discarding 16 portable tests that I verified compile and pass without `EOS_NET_POSIX`. No architectural deviation.

### Findings

|---|---|---|---|---|
| 1 | Medium | `tests/CMakeLists.txt:196-203` | `add_executable(test_net)` and `add_test(NAME test_net)` were unconditional; the diff moves both inside the new gate. On Windows the target disappears entirely, so 16 of the 19 cases are lost, not 3. The 16 non-POSIX cases (`net init/deinit`, socket/bind/listen/close argument validation, `resolve_null`, `send`/`recv`, `http_get`/`http_post`, `mqtt_connect`, `mdns_init`) use only `<stdio.h> <stdlib.h> <string.h> <assert.h>` plus project headers, and the `#ifdef EOS_NET_POSIX` block at `tests/test_net.c:171` already excludes every POSIX-only header and case. | Revert to the original structure and apply the same one-token change used in the root file — leave `add_executable`/`target_link_libraries(eos_net)`/`add_test` outside the gate and narrow only the inner block to `if(NOT CMAKE_CROSSCOMPILING AND NOT WIN32)`. Diff in **Proposed changes**. |
| 2 | Low | `tests/CMakeLists.txt:196` (side effect) | Same cause as #1, cross-compile path: a cross build with `-DEOS_BUILD_TESTS=ON` also loses the target, which contradicts the PR body's "All other platforms are unaffected since the added condition only narrows an existing gate." In `tests/CMakeLists.txt` the gate is newly applied, not narrowed. CI is unaffected — `.github/workflows/ci.yml:135` sets `-DEOS_BUILD_TESTS=OFF` for the cross job — but a local cross build with tests on is. | Fixed by the same diff as #1. Correct the claim in the PR body. |
| 3 | Low | PR body, "Changes" | "matching the existing pattern used for `test_pkg_trust_anchor`" — no target, test or file by that name exists anywhere under the working root (`grep -rn test_pkg_trust_anchor` over `/home/srpatcha/eos`, 0 hits; the only `WIN32` occurrences in either `CMakeLists.txt` are the two this PR adds). The cited precedent does not exist. | Drop the claim or cite the actual precedent. |
| 4 | Low | PR body, "Testing" | `[x] Unit tests pass (ctest --test-dir build --output-on-failure)` and `[x] Manual testing performed` on "MSVC / Visual Studio 2026 Build Tools" are asserted with no command output. Per the review brief §5 the unsupported claim is itself the finding — this is the one platform nothing in CI covers, so it is the claim that most needs evidence. | Paste the Windows `cmake` configure output and the `ctest` summary into the PR body. |
| 5 | Low | `.github/workflows/ci.yml`, `.github/workflows/release.yml:187-210` | Nothing prevents this regression from returning. `ci.yml` is `ubuntu-22.04` only. `release.yml`'s `windows-build` runs on `windows-latest` but configures without `-DEOS_BUILD_TESTS` (default `OFF`, `CMakeLists.txt:18`), so it exercises the root-`CMakeLists.txt` fix — and only at tag time, never on a PR — and never the `tests/` fix at all. | Add a `windows-latest` job to `ci.yml` that configures with `-DEOS_BUILD_TESTS=ON` and runs `ctest --no-tests=error`. Out of scope for this PR; worth a follow-up issue. |

Not a finding, recorded as correct: `eos_net` keeps `net/src/net.c` unconditionally (`CMakeLists.txt:188`), so dropping `net_posix.c` on Windows does not leave a sourceless target, and `net.c:25`'s `#ifndef EOS_NET_POSIX` fallback still defines every `eos_net_*` symbol. The library-side and test-side `EOS_NET_POSIX` gates remain textually identical, which is the invariant that matters given the define is `PRIVATE` to `eos_net`.

### Architecture conformance

Conforms. Build-configuration change only, confined to Tier 1 `eos` (master design §21). No new `#include`, link line or manifest entry, so §5.1 dependency direction is untouched — `eos_net` still depends only downward on `eos_hal` (`CMakeLists.txt:205`). No public header, API or ABI change, so the §23.2 EoS API / Driver API contracts are unaffected. The `ADR-014 point 3` reasoning quoted in the existing comment block (host OS maps eNet onto POSIX sockets) is preserved: Windows simply is not a POSIX host, and the freestanding fallback in `net.c` is what §13 and ADR-014 record as the Planned state for targets without an IP stack.

### Proposed changes

Smallest sequence that keeps everything building. Replace `tests/CMakeLists.txt:192-203` with:

```cmake
# --- test_net: Networking abstraction ---
add_executable(test_net test_net.c)
target_link_libraries(test_net PRIVATE eos_net)
# The POSIX round-trip cases need the same switch the library was built with,
# plus pthreads for the listener they connect to. NOT WIN32 for the same
# reason as the eos_net gate in the top-level CMakeLists.txt: a native
# Windows build is not cross-compiling but has no <arpa/inet.h> either.
if(NOT CMAKE_CROSSCOMPILING AND NOT WIN32)
    target_compile_definitions(test_net PRIVATE EOS_NET_POSIX=1)
    find_package(Threads REQUIRED)
    target_link_libraries(test_net PRIVATE Threads::Threads)
endif()
add_test(NAME test_net COMMAND test_net)
```

That is the original block with `AND NOT WIN32` added — the same single-token change already made in the root file, and consistent with the rationale the PR gives for it. Leave `CMakeLists.txt:200` exactly as the PR has it.

### Verification I ran

At the PR head (`git archive 5ed01576 | tar -x -C /tmp/eos-138`), Linux, gcc 12:

- `cmake -B build -DEOS_BUILD_TESTS=ON -DEOS_PRODUCT=vbox_test -DCMAKE_BUILD_TYPE=Debug` → rc 0; `build/tests/CTestTestfile.cmake:65` still registers `test_net`. **Linux is not regressed by this diff.**
- `cmake --build build --target test_net` → rc 0; `./build/tests/test_net` → rc 0, `=== ALL 19 NET TESTS PASSED ===`.
- Windows-equivalent code path, i.e. the same sources with `EOS_NET_POSIX` undefined:
  `gcc -o /tmp/test_net_nonposix tests/test_net.c net/src/net.c -Inet/include -Iinclude -Ihal/include -Ikernel/include -DEOS_ENABLE_NET=1` → rc 0, no warnings, no POSIX header pulled in; running it → `=== ALL 16 NET TESTS PASSED ===`. This is the evidence for finding #1: those 16 cases build and pass with no POSIX dependency, and this PR removes them from Windows.

### Not checked

- **Nothing was built or run on Windows or with MSVC.** No Windows host was available to this review. The 16-test result above is gcc on Linux with the POSIX path disabled, which proves the cases carry no POSIX dependency but does *not* prove MSVC accepts them. The PR's own Windows claim is therefore still unverified by anyone — see finding #4.
- The reported `C1083: Cannot open include file: 'netdb.h'` failure was not reproduced; it is consistent with the code (`CMAKE_CROSSCOMPILING` is `FALSE` natively on Windows, so `EOS_NET_POSIX=1` was defined and `tests/test_net.c:172` pulled in `<arpa/inet.h>`), but reproduction requires the MSVC host.
- CI: `checks.txt` in the bundle is empty. No check-run conclusions existed when this review's context was captured, roughly eight minutes after the PR was opened. I did not re-poll GitHub, so I make no statement about whether required checks now pass.
- Cross-compilation with `-DEOS_BUILD_TESTS=ON` was not configured; finding #2 is read from the diff, not executed.


---
*Automated architecture review of `5ed015763799` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

ak-sharvesh:
Thanks for the thorough review you're right, my original gate
over-corrected and dropped 16 valid non-POSIX cases unnecessarily. Applied
your exact proposed fix (moved add_executable/target_link_libraries/add_test
back outside the gate, narrowed only the POSIX-specific inner block).

Verified locally on Windows (MSVC / VS 2026 Build Tools): ctest now shows
test_net passing (all 19 cases), confirming your fix works there too.

One small note on finding #3 test_pkg_trust_anchor does exist in the repo
for me at tests/CMakeLists.txt:217-221, gated the same way. Not sure why it
didn't turn up on your end  possibly a different snapshot but happy to
double check if useful.

Will update the PR body with the corrected test evidence and drop the "all
other platforms unaffected" overstatement per finding #2.

srpatcha (commented):
## Review — eos#138 "Fix: eos_net/test_net fail to build on native Windows"
head: 761026b56e22173715cba8e4b0c86df8095efe5b  author: ak-sharvesh  ci: none (zero check runs on the branch — see finding 3)

**Verdict:** Re-review at the new head. The `tests/CMakeLists.txt` over-correction is fixed exactly as proposed — `add_executable`/`target_link_libraries`/`add_test` are back outside the gate and only the POSIX inner block is narrowed. Both halves of the diff are now correct and I verified Linux is not regressed. Findings 1 and 2 from the review of 5ed01576 are resolved. **My finding 3 on 5ed01576 was wrong and I withdraw it** — see below. What remains is Low and mostly about the PR body and about CI coverage, not the code.

### Findings

|---|---|---|---|---|
| 1 | Low | PR body, "Testing" | The pasted evidence reports `84% tests passed, 6 tests failed out of 38`, and the prose calls that "a clean build + `ctest` run". A run with six failing targets is not a clean build; the body should not describe it as one. The `test_net` line itself (`Test #30: test_net ... Passed 0.03 sec`) is exactly the evidence that was missing and it does answer finding 4 on the previous head. | Reword "clean build" to state what actually happened: configure succeeded, six unrelated targets fail on MSVC, `test_net` builds and passes. |
| 2 | Low | PR body, "Testing" | "The 6 failures are pre-existing and unrelated to this change" is asserted with no baseline. A true before/after is impossible here — before this patch the Windows configure/compile fails hard with C1083 — so the claim cannot be shown the usual way. The named targets (`test_os_adapter`, `test_config`, `test_hal`, `test_e2e`, `test_pkg_trust_anchor`, `test_stress`) do not link `eos_net` and the causes given (missing `unistd.h`, GCC-only `__attribute__`, unresolved symbols) are independent of this diff, so the claim is credible — it is just not evidenced. Note also that `test_pkg_trust_anchor` is a security test (`eos_pkg` must refuse to verify without a real trust anchor, tests/CMakeLists.txt:217-223); a failing security test should be filed, not absorbed into a footnote. | Say "believed pre-existing, not verified against a baseline because the pre-patch tree does not configure on Windows", and open a tracking issue for the six MSVC failures, calling out `test_pkg_trust_anchor` separately. |
| 3 | Low | .github/workflows/ci.yml:29,91,118,156,177 | Nothing stops this regression returning, and nothing has run on this PR. Every `ci.yml` job is `runs-on: ubuntu-22.04`; there is no Windows job. `release.yml`'s `windows-build` configures without `-DEOS_BUILD_TESTS`, which defaults OFF (CMakeLists.txt:18), so it exercises the root-`CMakeLists.txt` fix only, and only at tag time. The `tests/CMakeLists.txt` fix is covered by no automation at all. Separately, `ci.yml` does trigger on `pull_request` to `master` (`.github/workflows/ci.yml:6-7`) yet `statusCheckRollup` is empty and `gh pr checks 138` reports "no checks reported on the branch" — consistent with workflows held for maintainer approval on an outside contribution. So there is currently **no CI evidence of any kind** for this PR. | Out of scope for this PR: file a follow-up to add a `windows-latest` job to `ci.yml` configuring `-DEOS_BUILD_TESTS=ON` and running `ctest --no-tests=error`. In the meantime a maintainer needs to release the pending workflow runs before this merges. |

**Withdrawn — finding 3 on head 5ed01576 was incorrect.** I claimed `test_pkg_trust_anchor` did not exist anywhere in the working root. It does: `tests/CMakeLists.txt:217-223` at this head and at the previous one, plus `tests/test_pkg_trust_anchor.c`, `services/pkg/eos_pkg.{c,h}` and `CHANGELOG.md`. The author was right to push back, and the "matching the existing pattern" claim in the original PR body was accurate. My grep ran against a stale snapshot. The rest of that review stands.

**Not a finding, recorded as correct.** The two gates are still textually identical (`CMakeLists.txt:200`, `tests/CMakeLists.txt:198`), which is the invariant that matters because `EOS_NET_POSIX` is `PRIVATE` to `eos_net`. Dropping `net_posix.c` on Windows does not leave a sourceless target — `net/src/net.c` is unconditional (CMakeLists.txt:188) and its `#ifndef EOS_NET_POSIX` fallback still defines every `eos_net_*` symbol.

### Architecture conformance

Conforms. Build-configuration change only, confined to Tier 1 `eos` (master design §21). No new `#include`, link line or manifest entry, so §5.1 dependency direction is untouched — `eos_net` still depends only downward on `eos_hal` (CMakeLists.txt:205). No public header, API or ABI change, so the §23.2 EoS API and Driver API contracts are unaffected. The §13 eNet model is preserved: a host build maps eNet onto POSIX sockets, Windows is simply not a POSIX host, and the freestanding fallback in `net.c` is the Planned state §28 records for targets without an IP stack.

### Proposed changes

None to the code. The diff is what I asked for. Remaining work is editorial (findings 1-2) and a follow-up issue (finding 3).

### Verification I ran

At this head (`git archive 761026b5 | tar -x`), Linux, gcc 12:

- `cmake -B build -DEOS_BUILD_TESTS=ON -DEOS_PRODUCT=vbox_test -DCMAKE_BUILD_TYPE=Debug` → rc 0. `build/tests/CTestTestfile.cmake:65` still registers `test_net`, so **the target survives configure and Linux is not regressed**.
- `cmake --build build --target test_net` → rc 0; `./build/tests/test_net` → rc 0, `=== ALL 19 NET TESTS PASSED ===`.
- Windows-equivalent code path (same sources, `EOS_NET_POSIX` undefined): `gcc -o /tmp/tn tests/test_net.c net/src/net.c -Inet/include -Iinclude -Ihal/include -Ikernel/include -DEOS_ENABLE_NET=1` → rc 0, no warnings, no POSIX header reached; running it → `=== ALL 16 NET TESTS PASSED ===`. Those are the 16 cases the previous head deleted and this head keeps.
- `git grep -n trust_anchor 761026b5` → 5 files, including `tests/CMakeLists.txt:217`. This is what retracts my earlier finding 3.
- `gh pr checks 138` → "no checks reported on the 'fix/windows-net-build' branch"; `gh pr view 138 --json statusCheckRollup` → `[]`.

### Not checked

- **Nothing was built or run on Windows or with MSVC.** No Windows host was available. The 16-test result above is gcc on Linux with the POSIX path disabled: it proves those cases carry no POSIX dependency, it does not prove MSVC accepts them. The author's Windows `ctest` output is the only evidence for the MSVC half and it has not been independently reproduced.
- I did not verify that the six reported MSVC failures are pre-existing — see finding 2 for why a baseline is not obtainable from the pre-patch tree.
- I did not simulate `WIN32` through CMake; that the target now survives a Windows configure is read from the CMake text (`add_executable` and `add_test` sit outside the `if`), not executed.
- Cross-compilation with `-DEOS_BUILD_TESTS=ON` was not configured. It is fixed by the same change as the Windows path, but I did not run it.
- CI: no check run has executed, so I make no statement about whether required checks would pass.


---
*Automated architecture review of `761026b56e22` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

ak-sharvesh:
Appreciate you double-checking and being upfront about the stale grep no worries at all, these things happen. Updated the PR body per findings #1 and #2 to be more precise about the test evidence and the "pre-existing" claim. #3 (Windows CI coverage) sounds right as a follow-up issue rather than something blocking this PR happy to file that if useful.

ak-sharvesh:
Appreciate you double-checking and being upfront about the stale grep no
worries at all, these things happen. Updated the PR body per findings #1
and #2 to be more precise about the test evidence and the "pre-existing"
claim. #3 (Windows CI coverage) sounds right as a follow-up issue rather
than something blocking this PR happy to file that if useful.

srpatcha (approved):

ak-sharvesh:
Hi,

Just wanted to follow up PR #138 was reviewed, revised based on the
maintainer's feedback, and merged into master earlier today.

Thank you for the opportunity to work through this assessment, I enjoyed
digging into the codebase.

 Happy to discuss next steps whenever works for you.

Best,
Sharvesh

PR #139, opened 2026-09-04
fix(kernel): remove stale queue waiter entry on post-wakeup retry success

oscar-leung:
## Issue

A task that blocks in `eos_queue_send()` / `eos_queue_receive()` with a timeout enlists itself in the queue's waiter array. If its timeout fires but the queue gained capacity (or an item) before it retried, the retry succeeds and returns `EOS_KERN_OK` while the task's entry remains in the waiter list. The stale entry (1) leaks one of the 8 waiter slots and (2) is popped by the next wakeup, so `eos_task_unblock()` spuriously readies a task blocked on something else while the genuine waiter behind it loses its wakeup — hanging until its own timeout, or forever with `EOS_WAIT_FOREVER`. `eos_sem_wait()` in `kernel/src/sync.c` already dequeues itself on this exact retry-success path; `ipc.c` omitted it in both directions.

## Approach

Add the same "remove from wait queue if still there" dequeue to the retry-success paths of both `eos_queue_send` and `eos_queue_receive`, inside the existing critical section, matching the file's inline removal style. No API or behavior change on any other path.

## Testing / validation

- Static analysis of the waiter-list lifecycle across `ipc.c` / `sync.c` / `task.c`; both modified files pass `-fsyntax-only`.
- Added a deterministic host-side regression test (`test_queue_no_stale_send_waiter` in `tests/test_kernel.c`, driven through the existing yield hook): it reproduces the race and asserts the victim task is not spuriously unblocked. It fails before the fix and passes after.
- The test was written for CI to run; it was not executed locally.

## Limitations

Covers the queue primitives' send/receive paths; `eos_queue_delete`'s wake-all behavior is unchanged. The regression test exercises the send-side race directly and the receive-side fix by symmetry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kernel/src/ipc.c
+++ b/kernel/src/ipc.c
@@ -102,6 +102,16 @@ int eos_queue_send(eos_queue_handle_t h, const void *item, uint32_t timeout_ms)
         memcpy(g_q[h].buf + g_q[h].head * g_q[h].item_size, item, g_q[h].item_size);
         g_q[h].head = (g_q[h].head + 1) % g_q[h].capacity;
         g_q[h].count++;
+        /* Remove from wait queue if still there (woken by timeout, not by a
+         * receiver) — a stale entry would later steal another waiter's wakeup. */
+        for (int i = 0; i < g_q[h].send_waiter_count; i++) {
+            if (g_q[h].send_waiters[i] == caller) {
+                for (int j = i; j < g_q[h].send_waiter_count - 1; j++)
+                    g_q[h].send_waiters[j] = g_q[h].send_waiters[j + 1];
+                g_q[h].send_waiter_count--;
+                break;
+            }
+        }
         eos_port_exit_critical(crit);
         return EOS_KERN_OK;
     }
@@ -166,6 +176,16 @@ int eos_queue_receive(eos_queue_handle_t h, void *item, uint32_t timeout_ms)
         memcpy(item, g_q[h].buf + g_q[h].tail * g_q[h].item_size, g_q[h].item_size);
         g_q[h].tail = (g_q[h].tail + 1) % g_q[h].capacity;
         g_q[h].count--;
+        /* Remove from wait queue if still there (woken by timeout, not by a
+         * sender) — a stale entry would later steal another waiter's wakeup. */
+        for (int i = 0; i < g_q[h].recv_waiter_count; i++) {
+            if (g_q[h].recv_waiters[i] == caller) {
+                for (int j = i; j < g_q[h].recv_waiter_count - 1; j++)
+                    g_q[h].recv_waiters[j] = g_q[h].recv_waiters[j + 1];
+                g_q[h].recv_waiter_count--;
+                break;
+            }
+        }
         eos_port_exit_critical(crit);
         return EOS_KERN_OK;
     }
--- a/tests/test_kernel.c
+++ b/tests/test_kernel.c
@@ -187,6 +187,68 @@ static void test_queue_send_waiter_overflow(void) {
     printf("[PASS] queue send waiter overflow\n");
 }
 
+/* ---- Stale send-waiter after a timeout/wakeup race ---- */
+
+static eos_queue_handle_t q8_q;
+static eos_task_handle_t q8_a, q8_b, q8_r;
+static int q8_depth;
+
+static void q8_hook(void)
+{
+    q8_depth++;
+    if (q8_depth == 1) {
+        /* A is blocked in send_waiters. B piles in behind it. */
+        eos_task_set_current_internal(q8_b);
+        int item = 2;
+        assert(eos_queue_send(q8_q, &item, 10) == EOS_KERN_OK);
+    } else if (q8_depth == 2) {
+        /* Both A and B are queued as send waiters. A receiver drains one
+         * item: that pops A (head of the wait list) and frees the slot,
+         * which B's retry then consumes. B must leave the wait list on
+         * that success path — otherwise a stale entry for B remains. */
+        eos_task_set_current_internal(q8_r);
+        int out = 0;
+        assert(eos_queue_receive(q8_q, &out, EOS_NO_WAIT) == EOS_KERN_OK);
+        assert(out == 42);
+    }
+}
+
+static void test_queue_no_stale_send_waiter(void) {
+    eos_kernel_init();
+    assert(eos_queue_create(&q8_q, sizeof(int), 1) == EOS_KERN_OK);
+
+    int filler = 42;
+    assert(eos_queue_send(q8_q, &filler, EOS_NO_WAIT) == EOS_KERN_OK);
+
+    q8_a = (eos_task_handle_t)eos_task_create("q8a", test_entry, NULL, 5, 512);
+    q8_b = (eos_task_handle_t)eos_task_create("q8b", test_entry, NULL, 6, 512);
+    q8_r = (eos_task_handle_t)eos_task_create("q8r", test_entry, NULL, 7, 512);
+    q8_depth = 0;
+
+    q7_yield_hook = q8_hook;
+    eos_task_set_current_internal(q8_a);
+    int item = 1;
+    /* B stole the slot A was woken for, so A legitimately times out. */
+    assert(eos_queue_send(q8_q, &item, 10) == EOS_KERN_TIMEOUT);
+    q7_yield_hook = NULL;
+
+    /* B's send already succeeded; it is now blocked on something else
+     * entirely (here: a plain sleep). */
+    eos_task_block_with_timeout(q8_b, 1000);
+    assert(eos_task_get_state(q8_b) == EOS_TASK_BLOCKED);
+
+    /* Drain B's item. If B's entry went stale in send_waiters, this
+     * spuriously unblocks B from its unrelated sleep. */
+    eos_task_set_current_internal(q8_r);
+    int out = 0;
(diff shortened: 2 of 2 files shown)


srpatcha (commented):
## Review — eos#139 "fix(kernel): remove stale queue waiter entry on post-wakeup retry success"
head: 6a1971c0296586f2b3d842db4fe5c9fe167cfbf7  author: oscar-leung  ci: pending (no check runs in bundle; `mergeStateStatus: BLOCKED`)

**Verdict:** The bug is real, the fix is right, and I verified both by building and running the suite — but the retry-success path this PR is editing has a *second* omission of the same kind that the change does not address, and one of the tests guarding this exact code has never been called from `main()`.

Verification I ran (extracted `origin/master` with `git archive`, applied `diff.patch`, `cmake -B build/host -G Ninja -DCMAKE_BUILD_TYPE=Release -DEOS_BUILD_TESTS=ON`):

```
base ipc.c + this PR's new test:
  test_kernel: tests/test_kernel.c:246: test_queue_no_stale_send_waiter:
  Assertion `eos_task_get_state(q8_b) == EOS_TASK_BLOCKED' failed.   [exit 134]

patched tree:
  [PASS] queue send success clears wait-list entry
  === ALL KERNEL TESTS PASSED (13/13) ===                            [exit 0]
  full build: 143/143 targets, 0 errors
  ctest: 100% tests passed, 0 tests failed out of 38
```

So the PR body's "it fails before the fix and passes after" — which the body honestly labels as not executed locally — is now confirmed, and the fix regresses nothing else in the host suite.

### Findings
|---|---|---|---|---|
| 1 | High (P1) | `kernel/src/ipc.c:99-118`, `:173-192` | The retry-success paths still omit the counterpart wakeup that the first-try paths perform. `eos_queue_send` at `:69-76` wakes a `recv_waiters[0]` after enqueuing; the retry-success block at `:100-117` enqueues identically and wakes nobody. Symmetrically, `eos_queue_receive:143-150` wakes a send waiter after freeing a slot, and its retry-success block at `:174-191` does not. Concrete loss: capacity-1 queue, full; task S blocks in `send_waiters`; receiver R1 drains, waking S and emptying the queue; receiver R2 then calls `eos_queue_receive(q, &x, EOS_WAIT_FOREVER)` on the now-empty queue and enlists; S's retry sees `count < capacity`, enqueues, returns `EOS_KERN_OK` — and R2 is never unblocked despite an item sitting in the queue for it. That is a permanent hang, not a delayed one. The PR modelled the change on `eos_sem_wait` (`kernel/src/sync.c:337-352`), but the analogy does not carry: a semaphore's retry-success *consumes* a token and cannot enable another waiter, whereas a queue's retry-success enables the opposite side. Copying the dequeue without the wake is exactly the half-transfer that leaves the gap. | Extract the existing wake block into a file-local helper and call it from all four sites, e.g. `static void q_wake_one(uint8_t *w, uint8_t *n)` that pops index 0 and calls `eos_task_unblock`. Then `eos_queue_send`'s retry-success calls it on `recv_waiters` and `eos_queue_receive`'s on `send_waiters`, matching `:69` and `:143`. |
| 2 | Medium (P2) | `tests/test_kernel.c:159` | `test_queue_send_waiter_overflow` is defined and never called — `main()` does not invoke it, and it is the only occurrence of the name in the repo (`git grep` on `origin/master`). The compiler says so out loud in the build I ran: `tests/test_kernel.c:159:13: warning: 'test_queue_send_waiter_overflow' defined but not used [-Wunused-function]`. It covers the `Q_MAX_WAITERS` overflow path — one of the two failure modes this PR's own body cites ("leaks one of the 8 waiter slots") — so the slot-leak half of the fix has no executing test at all. The hand-maintained `(12/12)`/`(13/13)` counter in the final `printf` is what let this hide: the number tracks the calls someone remembered to add, not the tests that exist. | Add `test_queue_send_waiter_overflow();` to `main()` and make the banner `(14/14)`. Better, drop the hand-counted literal — a `static int g_tests_run;` incremented per `[PASS]` cannot silently disagree with reality. |
| 3 | Medium (P2) | `kernel/src/ipc.c:72,109,122,146,183,196`; `kernel/src/sync.c:114,234,344,356,382` | The "shift the waiter array down over index `i`" removal is now open-coded eleven times across the two kernel sync files — six of them in `ipc.c` after this PR adds the sixth. The defect this PR fixes *is* one of those copies having been forgotten, and finding 1 is a second one; the duplication is not stylistic, it is the mechanism by which these bugs get written. | `static void waiters_remove(uint8_t *w, uint8_t *n, uint8_t task)` and `static uint8_t waiters_pop_front(uint8_t *w, uint8_t *n)` in a shared kernel-internal header collapse all eleven. Do it as a separate commit — per `.ai/architect.md`, do not restructure and change behaviour together. |
| 4 | Low (P3) | `kernel/src/ipc.c:70` vs `kernel/src/sync.c:229-236,376-384` | Two blocking primitives in the same kernel use different wakeup disciplines and neither is documented. Queues wake `waiters[0]` — strict FIFO. `eos_sem_post` and the mutex release scan for the numerically lowest priority value and wake that waiter. On an RTOS the difference is observable as priority inversion on queue-mediated handoff. Not introduced by this PR and not its job to fix, but the master design is silent on it, so I have filed a proposal (see below) rather than only a finding. | Out of scope here. Decide the policy, document it in the `eos_queue_*` header comments, and make the two files agree. |
| 5 | Low (P3) | `tests/test_kernel.c` (EOF) | The file still ends without a newline (`\ No newline at end of file` in the diff), so every future patch to `main()` carries a spurious context line. | Add the trailing newline while `main()` is already being edited. |

### Architecture conformance
Conforms. Per master design §12.1, queues are EoS IPC Core — "Queues / Mailboxes / Shared memory / Events / signals" — so `kernel/src/ipc.c` in the `eos` repo is the right home, and §21 places `eos` in Tier 1 - Foundation. §5.1 is satisfied and I checked it rather than assumed it: the diff adds no `#include`, no link line and no manifest entry, and `ipc.c`'s existing includes (`eos/kernel.h`, `eos/kernel_internal.h`, `eos/arch.h`, `string.h`) are all at or below the kernel tier — nothing points up toward eAI, eNI, eBrowser, eOffice, health or aerospace. §6's "networking, databases, AI and distributed communication must not be kernel responsibilities" is untouched.

§6.1's "interrupt, timer and synchronization semantics must be documented and testable" is what this PR advances: it converts an untested waiter-lifecycle invariant into an executing regression test. Finding 4 is the part of §6.1 that remains unmet.

No §23.2 compatibility impact: `files.txt` confirms the diff touches only `kernel/src/ipc.c` and `tests/test_kernel.c` — nothing under `include/` or `kernel/include/`, no public signature, and `queue_t` is file-static to `ipc.c`, so no struct layout crosses a boundary. Nothing to declare, and the PR's silence on compatibility is correct here rather than an omission.

### Proposed changes
Smallest sequence that keeps the tree green:

1. **This PR, as is** — the dequeue fix is correct and verified. Merge it on its own merits.
2. **This PR or an immediate follow-up: finding 1.** Add the counterpart wake to both retry-success blocks. It is four lines each if written inline, or one helper plus four call sites. It needs a test; the existing `q7_yield_hook` harness cannot express it (see *Not checked*), so this likely needs the host port's `eos_port_yield()` to gain a way to keep a nested waiter blocked across the outer task's resume. That is worth doing once — the same harness limit will block every future lost-wakeup test.
3. **Follow-up commit: finding 2.** One line in `main()` plus the banner.
4. **Separate refactor commit: finding 3.** Behaviour-preserving, no test changes.

### Not checked
- **Finding 1 is Observed in the code, NOT reproduced.** I tried to build a host regression test for it and could not. The host harness's `eos_port_yield()` returns synchronously, so any task that enlists as a waiter inside the nested `q7_yield_hook` runs its own retry and dequeues itself before the outer task's retry resumes — there is no way to leave a genuine waiter enlisted across the retry. The reasoning above rests on reading `ipc.c:69-76` against `:100-117` and `:143-150` against `:174-191`, not on a failing assertion. Treat the severity as my judgement of the code, not as a measured hang.
- **ARM cross-compile: NOT RUN.** `.github/workflows/ci.yml` has a `Cross-compile ARM Cortex-M4` job; I built host x86-64 only (`gcc` 15.2.0 here, not the `gcc-12` CI pins). The diff is portable C with no arch-specific constructs, so I expect it to cross-compile, but I did not run it.
- **Python tests, cppcheck and clang-tidy: NOT RUN.** The `Run Python tests` and static-analysis legs of `ci.yml` were not executed here.
- **CI result: unknown.** `checks.txt` in the bundle is empty — the PR was opened minutes before this run and no check runs had reported. `mergeStateStatus` is `BLOCKED`, consistent with required checks still outstanding; I did not inspect branch protection.
- I ran the suite under `-DCMAKE_BUILD_TYPE=Release`, matching CI. `tests/CMakeLists.txt` adds `-UNDEBUG` for test targets only, so the `assert()`s were live — confirmed by the base-tree run aborting with exit 134 rather than passing silently.
- The local `/home/srpatcha/eos/eos` clone is 3 commits behind `origin/master`; every file I read came from `origin/master` via `git cat-file`/`git archive`, not the working copy.
- I did not review `eos_queue_delete`'s wake-all path, which the PR body correctly scopes out.


---
*Automated architecture review of `6a1971c02965` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #165, opened 2026-09-15
fix(os_services): OTA download and install run nothing through a shell

### LINKED ISSUE #164: OTA: eos_ota_download() and eos_ota_install() run attacker-controlled strings through system()
Found by the automated review of #156 (finding 1, Critical, pre-existing), which demonstrated both by execution.

## Where
`services/os/src/os_services.c`: `is_url_safe()` (`:83-87`), `eos_ota_download()` (`:103-107`), `eos_ota_install()` (`:127-133`).

## What
Both functions build a shell line with `snprintf` and hand it to `system()`, each behind its own denylist:

- `eos_ota_download()`: `wget -q -O "%s" "%s"` with `ota->local_path` and `ota->url`. `is_url_safe()` is `strpbrk(url, ";|><$()\`'")` — it omits `"`, `&`, `\`, newline and control characters. `eos_ota_set_source()` with `http://example.invalid/fw.bin" & touch /tmp/probe & "` followed by `eos_ota_download()`: **the sentinel is created**; the function then returns `-1` because `wget` fails, so the caller sees a failed download after the injected command has run.
- `eos_ota_install()`: `cp -f "%s" "%s"` with `target_path`. Its denylist `;|&><$()"'` includes `"` and `&` but omits the **backtick**, which is command substitution inside double quotes. `target_path = /tmp/x\`touch /tmp/probe\``: **sentinel created, return 0, `state = EOS_OTA_COMPLETE`** — an install that ran an attacker's command and reported success.
- `local_path` is unvalidated, and on Windows is derived from `getenv("TEMP")`.

This is the class #156 removed from the build backends and #119 removed from `linux_security`; the OTA path kept a private denylist of its own. `.ai/security.md` places OTA in the TCB.

## Expected
No shell between the caller's strings and the system: run the downloader with an argv (`posix_spawnp` / `_spawnvp`) and copy the file in C. Check the URL only for what the downloader would do with it — `http://`/`https://` and RFC 3986 characters, so it cannot be read as an option or a local file — and test both probes as data, asserting the sentinel is never created.

Kartikey1306:
Fixes #164

**Stacked on #144; review `9569226..fc3cb2f` for this change alone** -- three commits; `fc3cb2f` reads the installed image's mode off the open descriptor (`fstat`) rather than `stat`-ing the path the test then removes, which CodeQL flagged as check-then-use; `0b10524` clears the CodeQL check-run (two high alerts: the installed image and a test fixture were created with `fopen("wb")`'s 0666; both are `open(O_CREAT, 0600)` + `fdopen()` now). `990d862` alone was three files: `services/os/src/os_services.c`, `tests/test_os_services.c`, one CHANGELOG line at the top of `### Fixed`. Merge #144 first. This is #156's review finding 1 (Critical, pre-existing), done as the separate PR that review asked for; it does not depend on #156 because it does not need `EosShellCmd` -- there is no shell line left to build.

## The defect
`eos_ota_download()` and `eos_ota_install()` built `wget -q -O "%s" "%s"` and `cp -f "%s" "%s"` with `snprintf` and handed them to `system()`, each behind its own denylist with its own holes. The URL's omitted `"` and `&`: a URL of `http://example.invalid/fw.bin" & touch probe & "` left the quoting, ran the command, then reported a failed download. The target path's omitted the backtick, which is command substitution inside double quotes: a target of ``x`touch probe` `` ran the command and reported `EOS_OTA_COMPLETE`. Both were demonstrated by execution in review; the first is reproduced here by the new test's negative control.

## The fix
Neither step needs a shell, so neither uses one:
- the fetch runs `wget` (POSIX) or `curl` (Windows) with an **argv** through `posix_spawnp()` / `_spawnvp()`; nothing is parsed as shell syntax whatever the URL holds, and `local_path` (unvalidated before; from `TEMP` on Windows) is an argv element too;
- the install is a file copy in C (`fopen`/`fread`/`fwrite`, errors checked, `fclose` result checked).

The only check left on the URL is what the downloader would do with it: `url_is_fetchable()` requires `http://` or `https://` (so `-O…` cannot be read as an option and `file://…` is not a download) and RFC 3986 URL characters only (unreserved, gen-delims, sub-delims, `%`; no space, control, non-ASCII, `"`, `\`, backtick, `<>{}|^`). The sub-delims a signed URL carries -- `&`, `;`, `=`, `%`, `'` -- are allowed, because they were only ever dangerous to the shell that is no longer there. I chose the no-shell route over routing the two lines through `EosShellCmd` because `eos_shell_cmd_arg()` rightly refuses `&` and `;`, and an OTA URL with a query string is a normal thing to fetch.

## Verification
| Check | Result |
|---|---|
| `cmake -B build -DEOS_BUILD_TESTS=ON` → `cmake --build` | PASS, no warnings in the touched files |
| `ctest --test-dir build` | **43/43** |
| `tests/test_os_services` | **16/16** (12 + 4): the review's URL probe plus `-O…`, `file://`, `ftp://`, a backtick, a space, a control character and the empty string are all refused with `state == EOS_OTA_FAILED` and the sentinel absent; a URL with `?X-Sig=a%2Fb&Expires=1;v=2` passes the shape check (the fetch then starts, which is observable through `local_path` being set); the install copies a payload to a target named ``eos_ota_test_target`touch eos_ota_probe_install` `` byte for byte, `EOS_OTA_COMPLETE`, sentinel absent; a missing source, a missing `local_path`, `NULL` and `""` targets are refused |
| **Negative control:** `git show 9569226:services/os/src/os_services.c` restored, rebuilt, rerun | `test_ota_download_refuses_a_url_that_is_not_an_http_url` **fails at `tests/test_os_services.c:214: !file_exists(OTA_SENTINEL_DL)`** -- the injected `touch` ran |
| `pytest tests/unit -q` | 43 passed |
| Line endings | all three files LF on the base and LF here |

## Not done
- `eos_ota_verify()` still compares hex strings with `strcmp`; unchanged.
- The downloader choice (`wget` on POSIX) is as it was; a host without `wget` fails the fetch as before, now with `EOS_OTA_FAILED` from `posix_spawnp` rather than from a shell.
- #156's finding 2 (`eos_pkg.c` `pkill`/`rm -rf` lines fed from the package DB) is being handled on #163, which owns that file.

**CodeQL (check-run, not the Analyze job):** two high alerts on `990d862` -- `copy_file()` created the OTA target with `fopen(dst, "wb")` (0666 masked by umask) and the install test's fixture did the same. Cleared in `0b10524`: `open(O_WRONLY|O_CREAT|O_TRUNC, 0600)` + `fdopen()` behind a helper, `fopen` kept for the Windows branch. **0600 is deliberate for the image:** the updater writes it and reads it back for verification, so no other uid needs it, and firmware is not something other users should read; a consumer that needs it wider is a policy decision of its own. The install test asserts `(st_mode & 077) == 0` and was run under `umask 000`. Same class as `bba834a` and `8dca787`.


Kartikey1306:
CodeQL's check-run flagged two high alerts on `990d862` — "file created without restricting permissions" on `copy_file()`'s `fopen(dst, "wb")` and on the install test's fixture. Cleared in `0b10524` (range now `9569226..0b10524`); same class this repository fixed in `bba834a` and `8dca787`, same fix.

Both are `open(O_WRONLY|O_CREAT|O_TRUNC, 0600)` + `fdopen()` now, behind a small helper, with `fopen` kept for the Windows branch. For the installed image the mode is a deliberate choice rather than a default: the updater writes it and the updater reads it back for verification, so no other uid needs it, and firmware is not something other users should read — a consumer that needs it wider is a policy decision of its own, to be made where that consumer is. The install test asserts `(st_mode & 077) == 0` on the target, and I ran it under `umask 000` to show the mode does not depend on the environment. `ctest` 43/43, `test_os_services` 16/16.

Not touched: the two pre-existing `fopen("w")`/`"wb"` creations elsewhere in `os_services.c` (the audit log at `:285` and secure storage at `:370`) are outside this PR's lines; same class, worth their own pass.

Kartikey1306:
One more from CodeQL on `0b10524`'s own test: `stat(target)` followed by `remove(target)` is a check of a path and then a use of it. `fc3cb2f` reads the mode off the descriptor the test already holds open for the read-back (`fstat(fileno(f), ...)`) and nothing stats the path before it is removed. Range now `9569226..fc3cb2f`; `test_os_services` 16/16 under `umask 000`.

